PR #34, opened 2021-04-13
Ide capacitors

josephinekoe:


electronics_lib/Passives.py
```diff
@@ -313,6 +314,120 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      elif self.min_package_size not in self.DERATE_VOLTCO:
+        return (0, 0)
+      else:
+        voltco = self.DERATE_VOLTCO[self.min_package_size]
+        return (
+          capacitance[0] * (1 - voltco * (voltage[1] - 3.6)),
+          capacitance[1] * (1 - voltco * (voltage[0] - 3.6))
+        )
+
+    self.derated_capacitance = derated_capacitance()
```
ducky64, line 394: One thing to keep in mind: the algorithm here is going to look different than as implemented with the part table, since what that did is calculate the derated capacitance given the actual part (with set package and capacitance). Here, you're trying to calculate the package and nominal (label) capacitance given a capacitance spec (which we assume to be post-derating) - so it's kind of a reverse algorithm. Note that in this case, you may need to iterate up the package sizes - as you derate from a smaller package size, it may push up into a larger package size which has a less aggressive derating curve.

electronics_lib/Passives.py
```diff
@@ -313,6 +314,120 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      self.assign(self.selected_derated_capacitance, self.derated_capacitance)
+
+      self.footprint(
+        'C', self.footprint_spec,
+        {
+          '1': self.pos,
+          '2': self.neg,
+        },
+        value=f'{UnitUtils.num_to_prefix(value, 3)}F'
+        # TODO mfr, part, datasheet
```
ducky64, line 427: There isn't going to be a mfr, part number, or datasheet for a generic part

electronics_lib/test_capacitor_generic.py
```diff
@@ -0,0 +1,103 @@
...
+
+  def test_high_single_capacitor(self) -> None:
+    compiled = ScalaCompiler.compile(HighSingleCapacitorGenericTestTop)
+    self.assertEqual(compiled.get_value(['dut', 'footprint_name']), 'Capacitor_SMD:C_1206_3216Metric')
+    self.assertEqual(compiled.get_value(['dut', 'value']), '22uF')
+
+  def test_medium_single_capacitor(self) -> None:
+    compiled = ScalaCompiler.compile(MediumSingleCapacitorGenericTestTop)
+    self.assertEqual(compiled.get_value(['dut', 'footprint_name']), 'Capacitor_SMD:C_0805_2012Metric')
+    self.assertEqual(compiled.get_value(['dut', 'value']), '2.2uF')
```
ducky64, line 103: style: newline

electronics_lib/test_capacitor_generic.py
```diff
@@ -0,0 +1,103 @@
...
+    super().__init__()
+    self.dut = self.Block(SmtCeramicCapacitorGeneric(
+      capacitance=(50, 1000) * uFarad,
+      voltage=(0, 3.3) * Volt
+    ))
+    (self.dummya, ), _ = self.chain(self.dut.pos, self.Block(PassiveDummy()))
+    (self.dummyb, ), _ = self.chain(self.dut.neg, self.Block(PassiveDummy()))
+
+
+class HighVoltageCapacitorGenericTestTop(Block):
```
ducky64, line 34: I don't think you have voltage ratings as part of the cap selection process, right? If not, you can probably delete this.

ducky64 (changes requested):
Some suggestions inline, but I'm not sure the generic capacitor selection algorithm will do what we want it to do. I think if I were to sketch out the algorithm, it would look like:

Start with the smallest package given the spec capacitance (or the input footprint spec).
Get the derated (nominal / label) capacitance for that package given the capacitance spec.
Choose a post-derating preferred-number capacitance.
If the derated capacitance exceeds the package limits, go to the smallest package for that, and repeat.
If the input footprint spec isn't empty, check that it is satisfied (eg, it wasn't too small).
Somehow, work the capacitor splitting in here. Needs more thought on how specifically, but for simplicity you can assume you only split the largest package size.

Try adding the test case of a non-E24 capacitor spec and see if it'll snap to the closest E24. (practically, most capacitors especially in the uF range aren't E24, it's probably closer to E3 or E6, so we might want to add a control for that somewhere)

Let me know if this makes sense, we can discuss if it doesn't.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,224 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+  # 0805: 100v / 0.1uF,     50v / 0.1uF (maybe 0.22uF),     25v /  10uF
+  # 1206: 100v / 0.1uF,     50v /                4.7uF,     25v /  10uF,                      10v /  22uF
+  # 1210: 100v / 4.7uF,     50v /                 10uF,                      16v /  22uF,     10v /  47uF
+  # 1812: 100v / 2.2uF,     50v /                  1uF,     25v /  10uF (though small sample size)
+
+  # derating coefficients in terms of %capacitance / V over 3.6
+  # 'Capacitor_SMD:C_0603_1608Metric'  # not supported, should not generate below 1uF
+
+  PACKAGE_SPECS = {
+    402: {
```
ducky64: Style: consider using a [namedtuple](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#named-tuples) here so it's type-safe.

Also, consider having the package name be a string - it's only coincidence that the package names can (somewhat) be encoded as an integer. If you're curious, 0603 actually means 0.06" x 0.03".

electronics_lib/Passives.py
```diff
@@ -314,6 +297,224 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
 
+class SmtCeramicCapacitorGeneric(Capacitor, FootprintBlock, GeneratorBlock):
+  @init_in_parent
+  def __init__(self, *args, **kwargs):
+    super().__init__(*args, **kwargs)
+
+    self.footprint_spec = self.Parameter(StringExpr(""))
+
+    # Default to be overridden on a per-device basis
+    self.single_cap_max = 22e-6
```
ducky64, line 308: Should this be a `self.` field, or is this just a classwide (or even local function) constant?

electronics_lib/Passives.py
```diff
@@ -314,6 +297,224 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+    },
+  }
+
+  def select_capacitor_no_prod_table(self, capacitance: RangeVal, voltage: RangeVal,
+                                     single_nominal_capacitance: RangeVal,
+                                     part_spec: str, footprint_spec: str) -> None:
+
+    # capacitance: user-specified capacitance
+    # single nominal capacitance: no single cap with requested capacitance, must generate multiple parallel caps
+    # nominal capacitance: selected part's capacitance
```
ducky64, line 391: If this is meant to be a docstring, consider formatting as a docstring?

electronics_lib/Passives.py
```diff
@@ -314,6 +297,230 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+    ),
+    1210: SmtCeramicCapacitorGenericPackageSpecs(
+      name='Capacitor_SMD:C_1210_3225Metric',
+      max=float('inf'),
+      derate=0,
+      vc_pairs={100: 4.7e-6, 50:   1e-5,           16: 2.2e-5, 10: 4.7e-5},
+    ),
+    1812: SmtCeramicCapacitorGenericPackageSpecs(
+      name='Capacitor_SMD:C_1812_4532Metric',
+      max=float('inf'),
```
ducky64, line 379: There should still be some max here, for example you'd have a hard time finding a 1mF capacitor in even 1206 🙃

electronics_lib/Passives.py
```diff
@@ -314,6 +297,230 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+    402: SmtCeramicCapacitorGenericPackageSpecs(
+      name='Capacitor_SMD:C_0402_1105Metric',
+      max=1e-7,
+      derate=0,
+      vc_pairs={             50:   1e-7, 25: 1e-7,             10: 2.2e-6},
+    ),
+    603: SmtCeramicCapacitorGenericPackageSpecs(
+      name='Capacitor_SMD:C_0603_1608Metric',
+      max=1.1e-6,
+      derate=0,
```
ducky64, line 356: I think derating should be all-or-nothing: either it's applied and defined (even if kinda guessy) for all footprints, or none. If you find curves for 0603 / 0402, I'd recommend applying them; otherwise, consider removing it.

If you zero out the derating here, it may be worth keeping the code in case we use it in the future. However, would need to figure out how to test it.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,230 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+  # 1812: 100v / 2.2uF,     50v /                  1uF,     25v /  10uF (though small sample size)
+
+  # derating coefficients in terms of %capacitance / V over 3.6
+  # 'Capacitor_SMD:C_0603_1608Metric'  # not supported, should not generate below 1uF
+
+  class SmtCeramicCapacitorGenericPackageSpecs(NamedTuple):
+    name: str
+    max: float
+    derate: float
+    vc_pairs: dict
```
ducky64, line 344: Consider adding comments (or moving the above comments to here) that define these a bit better. I typically like adding units (eg footprint_name, max_capacitance) as well.

Specifically for max, may also be worth noting whether it's a nominal value or a maximum-after-tolerancing value. Note that when we do board design (manually) and parts selection, we typically work with nominal values.

ducky64 (changes requested):
Comments on data structures so far.

Working on reviewing the algorithm now.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+    # capacitance: user-specified capacitance
+    # single nominal capacitance: no single cap with requested capacitance, must generate multiple parallel caps
+    # nominal capacitance: selected part's capacitance
+
+    def valid_package_min_nominal_capacitance(package: int, capacitance: RangeVal, voltage: RangeVal) -> Tuple[float, float]:
+      """
+      Calculates if the package is a valid choice for the given capacitance and voltage.
+      If the package is a valid choice, the function returns the minimum nominal capacitance
+      for the specs. If the package is not valid, the function returns False.
+      """
```
ducky64, line 391: The type signature and description seem to disagree?

(looking at your code, I think you just need to update the type signature)

Also, consider returning an Optional[float] type - which can either be a float or `None`. I think that's the recommended pattern in this case.

You capacitance and voltage inputs might also be a Tuple[float, float] instead of a RangeVal.

ducky64 (changes requested):
Also, I think you should run the static type checker - the type signatures don't seem to match what's in the code.

See commands for running mypy in [developing.md](https://github.com/BerkeleyHCI/PolymorphicBlocks/blob/master/developing.md#static-checking)

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      return False
+
+    def select_package(capacitance: RangeVal, voltage: RangeVal) -> Tuple[float, float]:
+
+      if footprint_spec == "":
+        for package in sorted(self.PACKAGE_SPECS.keys()):
+          package_derated_capacitance = valid_package_min_nominal_capacitance(package, capacitance, voltage)
+          if package_derated_capacitance:
+            return (package, package_derated_capacitance)
+        return ("", (0, 0))
```
ducky64, line 419: Consider returning None in the case where it fails

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      else:
+        package_derated_capacitance = valid_package_min_nominal_capacitance(footprint_spec, capacitance, voltage)
+        if package_derated_capacitance:
+          return (footprint_spec, package_derated_capacitance)
+      return (footprint_spec, (0, 0))
```
ducky64, line 424: Can you unify this block with the case where `footprint_spec == ""`? One way I can think of is to create a package_options, which is `[footprint_spec]` (if `footprint_spec` is not empty) or `self.PACKAGE_SPECS.keys()` (if `footprint_spec` is empty)

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+
+    value = choose_preferred_number(capacitance, 0, self.E24_SERIES_ZIGZAG, 2)
```
ducky64, line 426: Value should be post-derating I think?
(for example, if a package with heavy derating was selected, the nominal capacitance would need to be significantly increased to get the desired effective capacitance)

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+        num_caps * min_nominal_capacitance[1],
+      ))
+
+      cap_model = DummyCapacitor(capacitance=(self.single_cap_max, self.single_cap_max),
+                                      voltage=self.voltage)
+      self.c = ElementDict[DummyCapacitor]()
+      for i in range(num_caps):
+        self.c[i] = self.Block(cap_model)
+        self.connect(self.c[i].pos, self.pos)
+        self.connect(self.c[i].neg, self.neg)
```
ducky64, line 448: Getting this to work properly with derating might actually be pretty tricky.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+        'C', valid_footprint_spec,
+        {
+          '1': self.pos,
+          '2': self.neg,
+        },
+        value=f'{UnitUtils.num_to_prefix(value, 3)}F'
+      )
+
+class DummyCapacitor(DummyDevice, Capacitor, FootprintBlock, GeneratorBlock):
+
```
ducky64, line 463: Is there supposed to be a newline inbetween the declaration and docstring?

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      if self.PACKAGE_SPECS[package].max >= avg_cap:
+        for v, c in self.PACKAGE_SPECS[package].vc_pairs.items():
+          if v >= voltage[1] and c >= min_nominal_capacitance[1]:
+            return min_nominal_capacitance
+      return False
+
+    def select_package(capacitance: RangeVal, voltage: RangeVal) -> Tuple[float, float]:
+
+      if footprint_spec == "":
+        for package in sorted(self.PACKAGE_SPECS.keys()):
```
ducky64, line 415: While sorting by name will work for the keys as currently exist, it might be brittle to change since it ties order of search with numerical value.
Instead, it might be preferable to have the packages be a list of PackageSpecs instead of a dict with a number as a key, and note in a comment that the list order is the search order.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      else:
+        voltco = self.PACKAGE_SPECS[package].derate
+        min_nominal_capacitance = (
+          capacitance[0] / (1 - voltco * (voltage[1] - 3.6)),
+          capacitance[1] / (1 - voltco * (voltage[0] - 3.6))
+        )
+
+      # Check if package is still valid for the derated capacitance
+      if self.PACKAGE_SPECS[package].max >= avg_cap:
+        for v, c in self.PACKAGE_SPECS[package].vc_pairs.items():
```
ducky64, line 407: consider more descriptive names, eg package_max_voltage, package_max_capacitance

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+      """
+      Calculates if the package is a valid choice for the given capacitance and voltage.
+      If the package is a valid choice, the function returns the minimum nominal capacitance
+      for the specs. If the package is not valid, the function returns False.
+      """
+      avg_cap = (capacitance[0] + capacitance[1]) / 2
+      avg_volt = (voltage[0] + voltage[1]) / 2
+
+      # Calculate derated capacitance for the package
+      if (avg_volt <= 3.6) or (avg_cap <= 1e-6):
```
ducky64, line 396: Consider using the maximum value instead of the average.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,222 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+    self.assign(self.selected_derated_capacitance, capacitance)
+
+    self.footprint(
+      'C', sorted(self.PACKAGE_MIN_CAP.keys())[-1], # TODO fix to a parameter footprint
+      {
+        '1': self.pos,
+        '2': self.neg,
+      },
+      value=f'{UnitUtils.num_to_prefix(capacitance[0], 3)}F'
+    )
```
ducky64, line 514: I think this could be vastly simplified by making it the responsibility of the calling function to select a package and capacitance, and all this does is create a footprint with the specified package and value without any additional logic.
(aside from perhaps creating the value string with proper sig figs in the footprint itself)

electronics_lib/test_capacitor_generic.py
```diff
@@ -0,0 +1,118 @@
...
+
+  def test_medium_single_capacitor(self) -> None:
+    compiled = ScalaCompiler.compile(MediumSingleCapacitorGenericTestTop)
+    self.assertEqual(compiled.get_value(['dut', 'footprint_name']), 'Capacitor_SMD:C_0805_2012Metric')
+    self.assertEqual(compiled.get_value(['dut', 'value']), '2.2uF')
+
+  def test_non_e24_capacitor(self) -> None:
+    compiled = ScalaCompiler.compile(NonE24CapacitorGenericTestTop)
+    self.assertEqual(compiled.get_value(['dut', 'footprint_name']), 'Capacitor_SMD:C_0603_1608Metric')
+    self.assertEqual(compiled.get_value(['dut', 'value']), '150nF')
```
ducky64, line 118: newline at end

electronics_lib/test_capacitor_generic.py
```diff
@@ -0,0 +1,118 @@
...
+    compiled = ScalaCompiler.compile(HighSingleCapacitorGenericTestTop)
+    self.assertEqual(compiled.get_value(['dut', 'footprint_name']), 'Capacitor_SMD:C_1210_3225Metric')
+    self.assertEqual(compiled.get_value(['dut', 'value']), '22uF')
+
+  def test_medium_single_capacitor(self) -> None:
+    compiled = ScalaCompiler.compile(MediumSingleCapacitorGenericTestTop)
+    self.assertEqual(compiled.get_value(['dut', 'footprint_name']), 'Capacitor_SMD:C_0805_2012Metric')
+    self.assertEqual(compiled.get_value(['dut', 'value']), '2.2uF')
+
+  def test_non_e24_capacitor(self) -> None:
```
ducky64, line 115: I'm planning to refactor out the non-E24 cases out of the resistor (it's a bad idea, it should be some alternative explicit opt-in mechanism instead), you can also just get rid of this now too.

ducky64 (changes requested):
Summary of comments:
- A bunch of style suggestions
- De-rating needs to affect the final capacitor value (I might not have been clear about this 🙃, see inline comments for more details). But because the derating tables are kind of spotty, and this can add significant complexity (especially when combined with the parallel capacitor generator), another option would just be to ignore derating completely. I'm actually leaning towards the latter. Or something simpler, like allowing the user to specify a derating factor via refinements. We can discuss this more in a meeting if you'd like.
- I suggest making DummyCapacitor even more simple, by just having it take in the footprint and value from its parent, and not doing any calculation in the DummyCapacitor class. See inline comment for more details.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+class SmtCeramicCapacitorGeneric(Capacitor, FootprintBlock, GeneratorBlock):
+
+  SINGLE_CAP_MAX = 22e-6
+
+  @init_in_parent
+  def __init__(self, *args, **kwargs):
+    super().__init__(*args, **kwargs)
+
+    self.footprint_spec = self.Parameter(StringExpr(""))
+    self.derating_coeff = self.Parameter(FloatExpr(1))
```
ducky64, line 309: May be useful to note in a comment here that this is capacitance derating, as a simple multiplier, and does not scale with package or applied voltage.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+
+    # Default to be overridden on a per-device basis
+    self.single_nominal_capacitance = self.Parameter(RangeExpr((0, self.SINGLE_CAP_MAX * 1.25)))  # maximum capacitance in a single part
+
+    self.generator(self.select_capacitor_no_prod_table, self.capacitance, self.voltage, self.single_nominal_capacitance,
+                   self.footprint_spec, self.derating_coeff)
+
+    # Output values
+    self.selected_capacitance = self.Parameter(RangeExpr())
+    self.selected_derated_capacitance = self.Parameter(RangeExpr())
```
ducky64, line 319: Might be clearer as selected_nominal_capacitance and selected_effective_capacitance?
Derating I think is more often used in talking about parts having a lower than labeled capacitance under load, whereas we're kind of applying the inverse here.

I'm also not sure the second is useful (pre-derating effective capacitance), because you'd need to then apply derating from the selected E-series and parallel caps to get the correct value.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+    min_nominal_capacitance = (capacitance[0] / derating_coeff, capacitance[1] / derating_coeff)
+
+    if min_nominal_capacitance[1] > single_nominal_capacitance[1]:
+      num_caps = math.ceil(min_nominal_capacitance[0] / self.SINGLE_CAP_MAX)
+      assert num_caps * self.SINGLE_CAP_MAX < min_nominal_capacitance[1], "can't generate parallel caps within max capacitance limit"
+
+      self.assign(self.selected_derated_capacitance, (
+        num_caps * capacitance[0],
+        num_caps * capacitance[1],
+      ))
```
ducky64, line 426: Are you sure about this? I'm not sure a meaningful number comes out if you multiply the original user-specified capacitance by the number of parallel capacitors?

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+
+      self.assign(self.selected_capacitance, (
+        num_caps * min_nominal_capacitance[0],
+        num_caps * min_nominal_capacitance[1],
+      ))
+
+      if footprint_spec == "":
+        split_package = 'Capacitor_SMD:C_1206_3216Metric'
```
ducky64, line 434: Consider having this defined similarly to and live next to SINGLE_CAP_MAX - these are quite inter-related.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+        if package.max >= min_nominal_capacitance[1]:
+            for package_max_voltage, package_max_capacitance in package.vc_pairs.items():
+              if package_max_voltage >= voltage[1] and package_max_capacitance >= min_nominal_capacitance[1]:
+                return package.name
+      return None
+
+    min_nominal_capacitance = (capacitance[0] / derating_coeff, capacitance[1] / derating_coeff)
+
+    if min_nominal_capacitance[1] > single_nominal_capacitance[1]:
+      num_caps = math.ceil(min_nominal_capacitance[0] / self.SINGLE_CAP_MAX)
```
ducky64, line 420: Why the use of both `single_nominal_capacitance` and `self.SINGLE_CAP_MAX`? Are these supposed to have some different meaning? I'd imagine you'd want to use `single_nominal_capacitance` everywhere to allow the user to override the maximum single capacitor value through the refinements system?

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+        package_options = [spec for spec in self.PACKAGE_SPECS if spec.name == footprint_spec]
+
+      for package in package_options:
+        if package.max >= min_nominal_capacitance[1]:
+            for package_max_voltage, package_max_capacitance in package.vc_pairs.items():
+              if package_max_voltage >= voltage[1] and package_max_capacitance >= min_nominal_capacitance[1]:
+                return package.name
+      return None
+
+    min_nominal_capacitance = (capacitance[0] / derating_coeff, capacitance[1] / derating_coeff)
```
ducky64, line 417: Naming: since it's an interval wouldn't this just be the nominal capacitance?

electronics_lib/Passives.py
```diff
@@ -314,6 +297,195 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+  @init_in_parent
+  def __init__(self, footprint_spec: StringLike = "", *args, **kwargs):
+    super().__init__(*args, **kwargs)
+
+    self.footprint_spec = self.Parameter(StringExpr(footprint_spec))
+    self.generator(self.select_capacitor, self.capacitance, self.footprint_spec)
+
+    # Output values
+    self.selected_capacitance = self.Parameter(RangeExpr())
+    self.selected_voltage_rating = self.Parameter(RangeExpr())
```
ducky64, line 476: Since there's no selection in this class / generator, can these just be deleted? (or in other words, would this data provide the user with any new information / insight?)

electronics_lib/test_capacitor_generic.py
```diff
@@ -0,0 +1,148 @@
...
+    self.assertEqual(compiled.get_value(['dut', 'c[0]', 'value']), '22uF')
+    self.assertEqual(compiled.get_value(['dut', 'c[1]', 'footprint_name']), 'Capacitor_SMD:C_1206_3216Metric')
+    self.assertEqual(compiled.get_value(['dut', 'c[1]', 'value']), '22uF')
+    self.assertEqual(compiled.get_value(['dut', 'c[2]', 'footprint_name']), 'Capacitor_SMD:C_1206_3216Metric')
+    self.assertEqual(compiled.get_value(['dut', 'c[2]', 'value']), '22uF')
+    self.assertEqual(compiled.get_value(['dut', 'c[3]', 'footprint_name']), 'Capacitor_SMD:C_1206_3216Metric')
+    self.assertEqual(compiled.get_value(['dut', 'c[3]', 'value']), '22uF')
+    self.assertEqual(compiled.get_value(['dut', 'c[4]', 'footprint_name']), 'Capacitor_SMD:C_1206_3216Metric')
+    self.assertEqual(compiled.get_value(['dut', 'c[4]', 'value']), '22uF')
+    self.assertEqual(compiled.get_value(['dut', 'c[5]', 'footprint_name']), None)
```
ducky64, line 148: end of file newline

ducky64 (changes requested):
Overall algorithm looks good, but some clean-up suggestions.

electronics_lib/Passives.py
```diff
@@ -314,6 +297,180 @@ def derated_capacitance(row: Dict[str, Any]) -> Tuple[float, float]:
...
+                                      voltage=self.voltage, footprint_spec=split_package)
+      self.c = ElementDict[DummyCapacitor]()
+      for i in range(num_caps):
+        self.c[i] = self.Block(cap_model)
+        self.connect(self.c[i].pos, self.pos)
+        self.connect(self.c[i].neg, self.neg)
+    else:
+      valid_footprint_spec = select_package(nominal_capacitance, voltage)
+      assert valid_footprint_spec is not None, "cannot generate a valid footprint spec"
+      value = choose_preferred_number(nominal_capacitance, 0, self.E24_SERIES_ZIGZAG, 2)
```
ducky64, line 439: I think the order of these should be flipped?
For example, it's been a common pattern to specify the upper range of capacitance at infinity - and I think as written, it wouldn't be able to find a package because it would look at the upper range?

However, if you choose a preferred number first, that would bound the range before the package selection logic runs.

ducky64 (changes requested):
Last thing (I hope). Everything else looks good!

ducky64 (approved):
Thanks, lgtm!

PR #36, opened 2021-05-06
Library dump utility

ducky64:
@rohit507 is this what you were looking for?

Behavior (with regards to what goes into superclass) may be incorrect before #35. The proto will also change after we add ordered KV pairs - but the code for this utility should not need to be updated.


rohit507 (approved):
Looks right to me, it'll be a while before I can properly test it. I'll need to finish the import code on my side first.

PR #37, opened 2021-05-12
BLE Parts

josephinekoe:


electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+      drain_voltage=(0, 4.5) * Volt
+    ))
+    self.co_fet = self.Block(NFet(
+      drain_current=(0, 4.5 / 150) * Amp,
+      power=RangeExpr.ZERO,
+      gate_voltage=(1.5, 3.4) * Volt,
+      gate_charge=RangeExpr.ALL,
+      rds_on=(0, 0.1) * Ohm,
+      drain_voltage=(0, 4.5) * Volt
+    ))
```
ducky64, line 62: Some comments about the specs:
- How was the drain current calculated - why `4.5 / 150`?
- Can you add a Default(...) around parameters that aren't derived from user specifications?
- Can the gate voltage be derived from the battery voltage? Same with the drain voltage?

electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+
+    self.ebp = self.Port(VoltageSource())
+    self.ebm = self.Port(GroundSource())
+    self.vdd = self.Export(self.battery_protector.vdd)
+    self.vss = self.Export(self.battery_protector.vss)
```
ducky64, line 67: Would it be clearer to use names like `pwr_in`, `pwr_out`, `gnd_in`, and `gnd_out`? (or perhaps `pwr_bat` / `pwr_out`?)

electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+
+  def contents(self) -> None:
+    super().contents()
+
+    # i/o connections
+    self.connect(self.vm_res.b.as_ground_source(), self.ebm)
+    self.connect(self.battery_protector.vdd, self.ebp)
```
ducky64, line 74: Doesn't this go through a resistor?

electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+    self.connect(self.vdd_res.b.as_voltage_sink(), self.vdd)
+
+    # vdd resistor
+    self.connect(self.battery_protector.vdd, self.vdd_res.a.as_voltage_sink())
```
ducky64, line 78: According to the example connection, I'd imagine the resistor terminal would be a source to the chip's sink?

electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+class BatteryProtector_S8200A(FootprintBlock):
+  @init_in_parent
+  def __init__(self, battery_voltage: RangeLike = RangeExpr()) -> None:
+    super().__init__()
+
+    self.battery_protector = self.Block(BatteryProtector_S8200A_Device())
+
+    self.vdd_res = self.Block(Resistor(resistance=330 * Ohm(tol=0.10)))
+    self.vm_res = self.Block(Resistor(resistance=2 * kOhm(tol=0.10)))
+    self.vdd_vss_cap = self.Block(Capacitor(capacitance=0.1 * uFarad(tol=0.10), voltage=battery_voltage))
```
ducky64, line 45: Consider using a DecouplingCapacitor, which is typed for power and ground ports.

electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+class BatteryProtector_S8200A(FootprintBlock):
+  @init_in_parent
+  def __init__(self, battery_voltage: RangeLike = RangeExpr()) -> None:
+    super().__init__()
+
+    self.battery_protector = self.Block(BatteryProtector_S8200A_Device())
+
+    self.vdd_res = self.Block(Resistor(resistance=330 * Ohm(tol=0.10)))
```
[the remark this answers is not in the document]
ducky64: Also take a look at `SeriesPowerResistor`, as used in the `Pcf2129` (realtime clock) class. Perhaps consider using the chain construct (also in that class) as well for readability?

ducky64 (changes requested):
Some primarily stylistic suggestions for the battery protector and battery class, along with two potential bugs.

(still working through the rest, but might as well post this part now)

electronics_lib/DcDcConverters.py
```diff
@@ -1,5 +1,85 @@
...
+        '3': self.vin,
+        '4': self.gnd,
+        '5': self.sw,
+        '6': self.vout,
+      },
+      mfr='Texas Instruments', part='TPS61023',
+      datasheet='https://www.ti.com/lit/ds/symlink/tps61023.pdf'
+    )
+
+class Tps61023(FootprintBlock):
```
ducky64, line 32: I recommend the `Ap3012` class as a modeling / starting point, and using the `DiscreteBoostConverter` superclass (but note one difference: the AP3012 is non-synchronous - it has a external blocking diode instead of an active switch that is integrated into the chip).

Also, note that the `Ap3012` uses a generator to size components from standard power converter equations, instead of using the suggested (generic) design. I'd also recommend using the generator approach here instead of static parameters - generators can better tailor a design to specific requirements. Can discuss more about power converters if you're interested.

electronics_lib/DcDcConverters.py
```diff
@@ -1,5 +1,85 @@
...
+    super().contents()
+    self.dcdc_converter = self.Block(Tps61023_Device())
+
+    self.vin_cap = self.Block(Capacitor())
+    self.vout_cap = self.Block(Capacitor())
+    self.ind = self.Block(Inductor())
+    self.vout_r1 = self.Block(Resistor())
+    self.vout_r2 = self.Block(Resistor())
+
+    self.footprint(
```
ducky64, line 53: Why does this have a footprint? Footprints should be physical components only, I don't think there's a 3-pin component that's part of this circuit?

ducky64 (changes requested):
Consider using the `DiscreteBoostConverter` superclass for the TPS60123 - it might be able to do a lot of work for you, and can automatically size parts to the application.

electronics_lib/Microcontroller_Adafruit_ItsyBitsy_BLE.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def pin_assign(self, pin_assigns_str: str) -> None:
+        system_pins: Dict[str, CircuitPort]
+        system_pins = {}
+
+        assigned_pins = PinAssignmentUtil(
+            AnyPinAssign([port for port in self._all_assignable_ios if isinstance(port, AnalogSink)],
+                         [0, 1, 2, 3, 4, 5]),
+            AnyPinAssign([port for port in self._all_assignable_ios if isinstance(port, DigitalBidir)],
+                         [0, 1, 2, 5, 7, 9, 10, 11, 12, 13]),
```
ducky64, line 85: Not sure this is correct - the list in the last argument should be the pin numbers on the chip. Note that the pin numbers on the chip correspond to a physical position (eg, pin 1 is top left, typically goes counterclockwise)), instead of the labeled pins.

ducky64 (changes requested):
Given that the BLE micro seems to be a very rough version (and doesn't seem to include the datasheet parameters, which would be the interesting and time-consuming part), what do you think about removing it (but keeping everything else - the boost converter and battery protection IC, but substituting out the micro in the example circuit so things are at least unit tested)?

If you were interested in building this physically, I could see a reason to continue, but I don't see myself continuing work for this chip since it's a bit limiting (for example, small number of pins, USB pins not available). I think BLE will still be interesting to play with some day, but perhaps with a different device.

The other stuff I still expect to be very useful, like the protection IC for lithium-ion batteries, and the boost converter (but for bringing a single NiMH / alkaline cell to 3.3v to power a micro).

electronics_lib/DcDcConverters.py
```diff
@@ -1,5 +1,81 @@
...
+class Tps61023(DiscreteBoostConverter, GeneratorBlock):
+
+  def contents(self):
+    super().contents()
+    self.assign(self.frequency, (0.5, 1)*MHertz)
+    self.assign(self.efficiency, (0.7, 0.97))
+
+    self.fb = self.Block(FeedbackVoltageDivider(
+      output_voltage=(580, 610) * mVolt,
+      impedance=(1, 300) * kOhm,
```
ducky64, line 41: I recommend setting the lower bound at 10k (unless the datasheet says otherwise?)
The main drawback of a lower impedance is higher power draw through the resistive divider. It will result in a 'stronger' signal, but past some point it doesn't matter and would just consume more power.

electronics_lib/DcDcConverters.py
```diff
@@ -1,5 +1,81 @@
...
+      datasheet='https://www.ti.com/lit/ds/symlink/tps61023.pdf'
+    )
+
+class Tps61023(DiscreteBoostConverter, GeneratorBlock):
+
+  def contents(self):
+    super().contents()
+    self.assign(self.frequency, (0.5, 1)*MHertz)
+    self.assign(self.efficiency, (0.7, 0.97))
+
```
ducky64, line 38: Can you also add a `require` on the output voltage here? It's 2.2-5.5v per the datasheet.

electronics_lib/DcDcConverters.py
```diff
@@ -1,5 +1,81 @@
...
+    self._generate_converter(self.ic.sw, 0.5,
+                             input_voltage=input_voltage, output_voltage=output_voltage,
+                             output_current_max=output_current[1], frequency=frequency,
+                             spec_output_ripple=spec_output_ripple, spec_input_ripple=spec_input_ripple,
+                             ripple_factor=ripple_factor)
+
+    self.connect(self.pwr_out, self.ic.vout.as_voltage_source(
+      voltage_out=self.pwr_out.voltage_out,  # TODO cyclic dependency?
+      current_limits=(0, 1.2)*Amp
+    ))
```
ducky64, line 78: Perhaps also a constraint on the effective inductance and capacitance range?
(this can be a todo - we don't have great support for this right now, but my understanding is that modern converters run with relatively high current ripple, and get away with smaller inductors)

electronics_lib/DcDcConverters.py
```diff
@@ -1,5 +1,81 @@
...
+
+    self._generate_converter(self.ic.sw, 0.5,
+                             input_voltage=input_voltage, output_voltage=output_voltage,
+                             output_current_max=output_current[1], frequency=frequency,
+                             spec_output_ripple=spec_output_ripple, spec_input_ripple=spec_input_ripple,
+                             ripple_factor=ripple_factor)
+
+    self.connect(self.pwr_out, self.ic.vout.as_voltage_source(
+      voltage_out=self.pwr_out.voltage_out,  # TODO cyclic dependency?
+      current_limits=(0, 1.2)*Amp
```
ducky64, line 77: This chip seems to have a higher current limit - "valley switch current limit" of 3.7A. You might also want to update the `rated_current_limit` in `_generate_converter`.

I'm not sure what valley switching current is, but if you have the interest see if you can understand how that translates into the EDG model, and if not, leave it as a TODO.

ducky64 (changes requested):
Looking much better! Still a few more issues.

electronics_lib/Batteries.py
```diff
@@ -45,3 +44,24 @@ def contents(self):
...
+
+  def contents(self):
+    super().contents()
+
+    # TODO can this be assigned self.pwr == VoltageSource(...) directly?
+    self.assign(self.pwr.current_limits, (0, 1)*Amp)
+    self.assign(self.capacity, (2, 3)*Amp)  # TODO should be A*h
+
+    self.footprint(
+      'U', 'Battery:BatteryHolder_Keystone_2461_1x57mm',
```
ducky64, line 61: Sadly this doesn't look like a KiCad part. But I think you can use `BatteryHolder_Keystone_2460_1xAA` instead - see the [online KiCad footprint list](https://kicad.github.io/footprints/Battery).

electronics_lib/Batteries.py
```diff
@@ -45,3 +44,24 @@ def contents(self):
...
+      {
+        '1': self.pwr,
+        '2': self.gnd,
+      },
+      mfr='Keystone', part='2461'
+    )
```
ducky64, line 67: Newline

electronics_lib/BatteryProtector_S8200A.py
```diff
@@ -0,0 +1,96 @@
...
+  @init_in_parent
+  def __init__(self) -> None:
+    super().__init__()
+
+    self.ic = self.Block(BatteryProtector_S8200A_Device())
+
+    self.pwr_out = self.Port(VoltageSource())
+    self.gnd_out = self.Port(GroundSource())
+    self.pwr_in = self.Port(VoltageSink(
+      voltage_limits=self.ic.vdd.voltage_limits,
```
ducky64, line 45: This actually is discarded, because the pin is connected (so it derives its properties from the inner connections.

If the system was a bit smarter about defaults processing, it would warn you if a non-default was specified.

electronics_lib/DcDcConverters.py
```diff
@@ -37,6 +119,7 @@ def contents(self):
     super().contents()
 
     self.require(self.pwr_out.voltage_out.within((0.76, 17)*Volt))
+    self.require(self.pwr_out.voltage_out.lower() <= self.pwr_in.voltage_limits.lower())
```
ducky64, line 122: I think here it's output voltage upper <= input voltage upper

electronics_lib/DcDcConverters.py
```diff
@@ -135,6 +218,8 @@ class Tps54202h(DiscreteBuckConverter, GeneratorBlock):
   def contents(self):
     super().contents()
 
+    self.require(self.pwr_out.voltage_out.lower() <= self.pwr_in.voltage_limits.lower())
```
ducky64, line 221: similarly, use upper

electronics_lib/DcDcConverters.py
```diff
@@ -253,6 +338,7 @@ def contents(self):
     super().contents()
 
     self.require(self.pwr_out.voltage_out.within((1, 24)*Volt))
+    self.require(self.pwr_out.voltage_out.lower() <= self.pwr_in.voltage_limits.lower())
```
ducky64, line 341: similarly, use upper

examples/test_AA_battery.py
```diff
@@ -0,0 +1,36 @@
...
+from edg import *
+from electronics_lib.Batteries import AABattery
+from electronics_lib.DcDcConverters import Tps61023
+
+
+class AABatteryCircuit(SimpleBoardTop):
+  def contents(self) -> None:
+    super().contents()
+    self.mcu = self.Block(Stm32f103_48())
+    self.li_ion_bat = self.Block(AABattery())
```
ducky64, line 12: It's not lithium here, perhaps just `bat`?

ducky64 (changes requested):
A few more minor things. Almost there!

ducky64 (approved):
Looks good, thanks!

(and that indeed turned out to be the last round of revisions!)

Issue #31, opened 2021-02-21
Clean up exporting of unconnected ports


ducky64:
Anything with the LCD's MISO not-connected causes compiler errors. Datalogger and high-side switch (lights controller) also has issues with unconnected CAN block pins, though is hacked around with explicit dummy block connections.

Solutions might include non-local compiler operations, or generating dummy blocks (either in the frontend or compiler).

ducky64:
New solution sketch: explicit not-connected - which is analogous to NC flags in schematics.

Implementation in IR: dummy block with port initialized to nop-defaults. Possibly some TBD syntactic sugar to generate these blocks - either as part of the port (eg, `DigitalBidir.makeDummy()`), or with infrastructural support (`DummyBlock(DigitalBidir)` - but where would this block definition live, automatically generated for all ports, for some ports, explicitly, or ???).

Alternatives considered:
- Remove unused exports - but requires chasing through the export and link chain, and would require deleting (or ignoring) elements from the IR that are not local to the block with the disconnected port. Likely difficult to implement and the result can be confusing.
- Automatically generate dummy blocks: requires core compiler support / refactoring (more complexity in an already somewhat-complex system), also wouldn't catch unintentionally disconnected ports (one of the more useful aspects of schematic ERC).

PR #39, opened 2021-08-30
Add hull operator, remove min(array[range]) -> float

ducky64:
- Add the hull reduction operator as the proper replacement for removing the min reduction operator on arrays of ranges, for calculating the superset link voltage given a set of bidirectional ports (and the optional source or sink port)
- Undefine (should error out) the min / max reduction on arrays of ranges - no longer used
- Undefine the min / max on empty array - no longer used
- Adds a hull binary operator (primarily because we don't have an array construction primitive - perhaps another refactor for another day)
- Add a proper concept of empty ranges (empty set equivalent for intervals) internal to the compiler
  - Split RangeValue into RangeType, RangeValue(min, max) and RangeEmpty
  - Hulling an empty array returns the empty range
  - Intersecting an empty array returns the full set - with the rationale being that the implicit initial value to intersect is the full set
  - @rohit507 is the overall behavior (primarily in ExprEvaluate.scala) consistent with formal definitions of empty intervals?
- Should not create any new test failures (but tests are currently failing from #31, unconnected ports causing missing values in assertions).

### DIFF
--- a/compiler/src/main/scala/edg/compiler/ConstProp.scala
+++ b/compiler/src/main/scala/edg/compiler/ConstProp.scala
@@ -96,7 +96,7 @@ class ConstProp {
       case init.ValInit.Val.Integer(_) => classOf[IntValue]
       case init.ValInit.Val.Boolean(_) => classOf[BooleanValue]
       case init.ValInit.Val.Text(_) => classOf[TextValue]
-      case init.ValInit.Val.Range(_) => classOf[RangeValue]
+      case init.ValInit.Val.Range(_) => classOf[RangeType]
       case _ => throw new NotImplementedError(s"Unknown param declaration / init $decl")
     }
     paramTypes.put(target, paramType)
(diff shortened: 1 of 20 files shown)


rohit507 (approved):
Looks good to me. 

Traditionally `RangeEmpty` would be any range where `min >= max` with equality testing being `(is_empty(l) && is_empty(r)) || ((lmin == rmin) && (lmax == rmax))` (and `is_empty(min,max): max <= min`). With the default empty range being `Range{min = inf, max = -inf}`.

Having an explicit empty constructor works identically, usually at the cost of more cases in the definition of the other operations. On the other hand you have to be careful to use non-overflowing operations on bounded types with the version above, which is a much harder error to deal with.

The benefits/drawbacks of each case are marginal, so do whatever's convenient.

ducky64:
Yeah, so the idea here of having a `RangeEmpty` is to avoid programming mistakes by requiring all instances of `RangeEmpty` to be handled explicitly (unless the min >= max solution provides unified and consistent logic across operators with the standard case?). There is an assertion guard on `RangeValue` construction requiring min <= max.

PR #40, opened 2021-09-21
Replace gRPC with stdin/stdout

### LINKED ISSUE #27: Less heuristic reload
Currently, it reloads everything by hooking imports to get the order of imports, along with a very heuristic ignore-list of system modules and core modules.

A better and more efficient solution would be to get the dependency tree from a specific module to the core package, but inspect.getmembers() and dir() both don't seem to be accurate for all of a module's dependencies.

ducky64:
Sockets cause weirdness (can only have one of them, might be phantom sockets if things aren't closed cleanly), and there doesn't seem to be an obvious way to have gRPC communicate over stdio. So this removes gRPC and replaces it with stdin / stdout between the host process and child process, using delimited protobuf (part of the library in Java / Scala, and custom implemented in Python).

Printf messages now go over stderr, to not interfere with the protobuf stream over stdout.

The HdlInterfaceServer also auto-starts when requested, given the filename to the script. This also no longer discards modules (resolves #27 by eliminating the problematic feature entirely!), since the code for that was extremely unreliable. Instead, this needs to be invoked anew when source files have changed.

Also fixes some (non-functional) type errors in top-level files, and adds Mypy IntelliJ plugin to developing guide.


rohit507 (approved):
Yes please.

PR #42, opened 2021-09-21
Add descriptive text proto

ducky64:
Protobuf infrastructure to support user-defined descriptive text for elements, which is structured as any combination of static strings and ValueExprs to be spliced into the strings. Includes both a short description (to be dumped in the block diagram) and a long description (eg, in a tooltip).

The long description would support HTML tags (which are supported by Swing tooltips without additional work) and should be composable (can include the description of a sub-element)

Open issues:
- Short descriptions might not be composable, since it should include the type name, and we wouldn't want a huge chain of wrappers (eg, `DecouplingCap(MlccCapacitor(0.1uF))` - instead we'd just want `DecouplingCap 0.1uF`). Maybe the short description should be a "value" instead (very loosely defined)?
- It might make sense to include short descriptions of sub-elements, instead of long descriptions. How should this be controlled? Perhaps the short description should be the default, and the long description as a separate ref.ValueExpr entry? Or perhaps have two description fields (long and short, as LocalPaths) separate from the value-splicing fields (because they're all quite different)?
- If a "value" field (instead of a short description field): how to get the full short description (that would include the type name)? Perhaps also a type name field? Or a read-only short description? Or a read-write short description that by default (in the frontend) just concats the type name field and value field?
- How this looks in the Python frontend is a separate question, but probably will be a function to be overridden that returns a List[Union[str, Elem, Parameter]]. If separating short and long description, we can probably have `Elem.short_desc` / `Elem.value` / `Elem.long_desc` accessors.


rohit507 (approved):
Seems good, I can just ignore the additions in my work.

ducky64:
Superseded by #125

PR #43, opened 2021-09-22
Clean up deprecations

ducky64:
DO NOT MERGE - DEPENDS ON #40 

... and add deprecations to the compiler options


rohit507 (approved):
Seems fine.

Issue #23, opened 2020-12-24
Abstract microcontroller type


ducky64:
One that has a grab-bag of common IO types (digital, SPI, I2C, UART, CAN, USB), using #20.

A future feature would allow the IOs to be of all of the above types, though the design model and implementation for that is unclear. For now, there would be an API similar to the current new_io that returns a io of the appropriate type. Main benefit is that it could allow micros to be more interchangeable with abstract types instead of changing HDL.

Related: #15, which is also relevant to this.

ducky64:
Thoughts on architecture of the components involved:

Chip level:
- instantiates the footprint
- ports more or less directly correlate with the footprint, probably configured as a named vector that correlates with the port name
- perhaps also creates peripheral ports (eg, SPI, I2C), but it's unclear how to handle configuration (either pin matrix like LPC1549 or alternative-mapping like STM32) and overlap with GPIO ports (perhaps mutual exclusion functionality? or pin mapping generator?)
  - peripherals at this level more accurately models the chip and its resources
  - this better modeling may also make it easier to generate firmware stubs (in the future), without needing things like graph traversal (but this will need to traverse the graph to get names - perhaps using link name with name inference from allocated port names)
  - otherwise, ports would need to be passively-typed, since analog functionality is often overlapped with digital ports

Application circuit level:
- contains required supporting components, like decoupling capacitors
- exposes the same interface as the chip above, except for anything only connecting to required components
  - requires vector export functionality
- does not contain optional components, like crystals (if there is an internal frequency source - but exposes the port)

Pin mapping level (IO controller wrapper / adapter):
- a very magical block, does a lot of things to present this software-controlled bag-of-IOs abstraction
  - everything else needed to use the microcontroller is automatically generated - for example, crystals (details below) and programming ports
  - parameters and refinements can be used to turn these off and control their implementation
- can contain mixins (#104) to provide the interface to support different functionality, eg frequency tolerance spec to automatically generate a crystal if needed
- contains a very basic pin mapper (greedy assignment - maybe better or configurable in the future), it is expected users will manually specify pin mappings through suggested name on the connected ports and refinements
  - if the chip or application circuit provides peripheral IOs: there are now overlapping responsibilities, in which case this will need to both handle direct port assignment (even bundles like SPI aren't broken up, they are directly forwarded) AND generating refinements to the chip to control its internal assignment
[...]

PR #46, opened 2021-09-25
Add not-connected port support, fix (most) red tests

### LINKED ISSUE #31: Clean up exporting of unconnected ports

ducky64:
DO NOT MERGE - DEPENDS ON #40, #43

Resolves #31, the two failing tests with the LCDs are now green. (Simon is still failing)

- Adds a `NotConnectablePort` mixin for `Port`, which adds the `not_connected()` function. This generates a dummy block (defined with the mixin) that defines a nop port for the port type, and connects the port there.
  - This is mainly (only?) intended to be used where there are unused bundle components - here, MISO is not used for the LCD
  - `not_connected()` is used to mark an edge port as disconnected, from the inside.
  - This differs from schematic capture, which marks pins not-connected from the outside. That functionality is handled by the required port infrastructure, and there is no way to NC a port from the outside. However, schematics have no concept of Bundles, where a sub-port is not-connected.
- Refactor the MCU pin assignment util to handle NC'd ports, which are `.not_connected()` at the top-level.
  - Architecturally this isn't the best, but should be cleaned by a separate refactor for #23

To some degree this is a bit of a hack, where the proper solution is better defined port types that don't include extraneous types. But defining those subset port types would be a much larger refactor job (including to-be-defined semantics), and this is probably a reasonable stopgap solution.

### DIFF
--- a/compiler/src/main/scala/edg/wir/DesignPath.scala
+++ b/compiler/src/main/scala/edg/wir/DesignPath.scala
@@ -96,7 +96,6 @@ object IndirectDesignPath {
 
 /**
   * Absolute path (from the design root) to some element.
-  * TODO: should exclude link ports, since the block side port is treated as authoritative.
   */
 case class DesignPath(steps: Seq[String]) {
   // Separates into (prefix, last) where last is the last element, and prefix is a DesignPath of all but the last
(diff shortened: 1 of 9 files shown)


rohit507 (approved):
Seems fine

Issue #49, opened 2021-09-27
Array operations


ducky64:
Use case: calculating parallel impedances, which needs 1/(1/x1 + 1/x2 + 1/x3 + ...).

While specialized operations (like SUM_INVERSE to define that particular reduction) can fix this case, the cleaner solution may be to define array operation.
Consistent with the behavior of numpy, define scalar-array operations - for example, 1/[1, 2, 3] = [1/1, 1/2, 1/3]. Also, define array-array operations, for example [1, 2, 3] / [4, 5, 6] = [1/4, 2/5, 3/6]

Note: [MATLAB distinguishes array vs. matrix operations](https://www.mathworks.com/help/matlab/matlab_prog/array-vs-matrix-operations.html), and has a dot prefix for element-wise array operations (where different - for example, + behaves the same in the matrix case as the elementwise case). But, as a Python HDL and not really supporting matrix data types yet, this probably isn't necessary yet. If needed in the future, the matrix case can be a different operator.

Other array operators, like dot-product, can be defined as separate operations.

IR definitions:
- Separate out scalar-scalar operations (eg, `ADD`), scalar-array operations (eg, `ADD_BROADCAST` - to use a [numpy term](https://numpy.org/doc/stable/user/basics.broadcasting.html)), and array-array operations (eg, `ADD_ELEMENTWISE`)

ducky64:
Resolved by #90 (kinda - in the compiler core with limited frontend support)

Issue #47, opened 2021-09-25
Clean up support for parallel analog sinks


ducky64:
This may be contributing to the Simon test failure (in addition to the exported speaker port, but with a resistor tacked onto one of the element ports, which isn't being handled properly - or really at all).

Currently, the solution defines conductance (really hackily) because of the lack of lambdas in the IR to do 1/x1 + 1/x2 + 1/x3...

Proposal: remove the hacktastic conductance (leaving only impedance), and define a SUM_INVERSE reduction in the IR to do 1/x1 + 1/x2 + 1/x3 + ...

PR #50, opened 2021-09-29
Simplify ConstraintExpr type parameters

ducky64:
Started yak shaving in working on #47 and #49.

This gets rid of the self-type and castable-type type parameters in the ConstraintExpr base class. The NumLike retains a castable type, because that's common code for the FloatExpr (to handle ints) and RangeExpr (to handle the all the float-like inputs) cases.

This mostly is changes to function signatures, except for a dynamic early conversion in the RangeExpr initializer.


rohit507 (approved):
Seems reasonable.

PR #52, opened 2021-10-04
Don't crash on empty designs

ducky64:
Allow namespace_order to handle the empty case (for which it returns an empty list), upper layers should crash if needed (for example, if there is a length mismatch between ordering).

Longer term: maybe eliminate the (unordered) map in the proto def so we don't need a supplemental order data structure?


rohit507 (approved):
Ok.

PR #53, opened 2021-10-04
Fix Analog link definition, fix Simon

ducky64:
wiiiiiiip


ducky64:
What do you mean fragile? Like not all cases are defined, or -0?

I'm not too worried about empty parallel impedances - it shouldn't be possible to have an analog link with no sink ports. But I think there's going to be a decent degree of building out more complete expression evaluation concepts as we have more motivating examples - because without those it's unclear whether we're building the right thing.

I think currently the reduce-on-empty definitions are 'good enough', but in the future it might be useful to allow explicit specification of either the initial value to be passed into a reduction, or the value returned on empty.

PR #59, opened 2021-10-08
New parts table

### LINKED ISSUE #3: Redo ProductTableUtils
It's obviously a dumpster fire. Ideas:

- Intervals as library
- Should take in arbitrary functions instead of composing table accessors, it becomes a parens nightmare
  - But perhaps there is room for higher order functions, eg ValueToRange(units=..., limit=zero/lower/upper, strict=T|F, missing=discard)
- Can there be more use of static typing in col accessors?
- Table methods: 
  - ProductTable(header, rows), ProductTable.from_csv(csvs)
  - duplicate (table is fundamentally mutable?)
  - map_derived(src_name, dst_name, fn: row_elt -> row_elt)
  - filter(row_dictlike)
  - sort_by_row(src_name) - ideally future more generalized sort function
  - first() -> row_dictlike

Migration ref: 89

ducky64:
Resolves #3

A new parts table that takes in standard functions instead of some odd DSL. This ended up pulling in a bunch of refactoring, but the main new code for this PR are in PartsTable.py and PassiveInductor.py

Changes:
- Add new parts table infrastructure, in electronics_lib/PartsTable.py
  - Add a LazyTable wrapper, that only loads the table when requested, but keeps the loaded table for future use
- Refactor inductor part table in electronics_lib/PassiveInductor.py, to use the new infrastructure, as a proof-of-concept
- Refactor mega-file Passives.py into resistor, capacitor, and inductor files
- Refactor out the E series utils into own utility class, separate from the part hierarchy
- Update README

A future PR will transition over the other part tables


rohit507 (approved):
Are there any plans to remove the dependency on the Digikey tables? Or at least allow for an alternate set of synthetic tables? 

Asking since this seems like a step in that direction and it would be nice for reproducibility as a research artifact.

ducky64:
Yeah, needing to download custom tables is a known major barrier to use. Ideas are:
- There's already an alternative rules-based capacitor table (maximum capacitance / voltage for footprint) that creates generic components - much like how the resistor generator works. Perhaps this just needs to be made the default.
- JLCPCB (which offers inexpensive low volume assembly services) has a full parts table available and doesn't appear to have any ToS nonsense, so there's a current work-in-progress to parse components from there. Downside is it's not fully parametric, so data needs to be parsed from the description field. This is mainly for passives, so would be another option for the rules-based options above.
- Not sure how to deal with discrete semiconductors like diodes and FETs, since I don't think a rules-based approach is feasible. I think a feasible option would be to create our own parts tables by manually transcribing a few datasheets, probably 10-100 components. Problem would be selecting a representative set of parts to use - would need a table of common part numbers, and we could then augment with parametric data. Maybe a recent version of the Art of Electronics textbook? The second edition in lab doesn't seem to have modern parts, but there is a third edition.

rohit507:
Generics for the passives should be the default. There's very little value
gained from a supplier database there.

Why wouldn't a generic approach work for diodes? They're standardized in a
convenient fashion, right?

ducky64:
The one advantage of a supplier database is that it makes ordering easier, since you don't need to translate from a spec (eg, 0.1uF, 25V) to a part number by doing a manual parametric search. Somewhat mitigated if you have a RC book or a pile of standard parts stocked, since you can directly work with those specs. But probably very useful for assembly services, where you need to give a part number.

Diodes: I don't think there's a standard? Even with resistors and capacitors, I don't think there's a formal standard, the rules are mostly from "for this capacitance and voltage rating, most of the parts are available in 0603 or 0805 or whatever". The package series for capacitors and resistors is also much more limited (primarily 0402/0603/0805/1206), but there's a larger selection of packages for diodes. There are also more parameters for diodes.

PR #61, opened 2021-10-09
Range type

ducky64:
Adds a Range type, for interval arithmetic, and including unit tests. Core definition is in Range.py, and refactors the code that uses tuples to use this. It's really mostly refactoring 🤪...

Additionally modifies the power converters dutycycle limits to be a constraint (instead of an assertion) so a design still generates, and a user may waive the constraint if it's fine for the converter to go into tracking mode. Fixes the battery protector example by removing the voltage-sensitive MCU and upping the output to 5V - previously it would have tracked up to the battery voltage of 4.2 and probably blew up the micro.

Prerequisite for #59, to replace the hacktastic old parts table adhoc range constructors.

TODO:
- [x] Refactor RangeExpr to use this internally
- [x] Refactor libraries to use this (mainly generators)
- [x] Define range operators as needed be examples
- [x] Clean up
- [x] Double-check the switching converter calculations
- [x] Netlist checks - spot checks seem fine, while the datalogger changes the converter values, it seems to match current master

### DIFF
--- a/edg_core/Binding.py
+++ b/edg_core/Binding.py
@@ -8,6 +8,7 @@ from typing import *
 from . import edgir
 from .Core import Refable
 from .IdentityDict import IdentityDict
+from .Range import Range
 
 if TYPE_CHECKING:
   from .Ports import BasePort, Port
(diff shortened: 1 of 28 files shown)


ducky64:
Did a bunch of cleanup and checking, should be good now.

Issue #16, opened 2020-10-29
Resistor generator fixes


ducky64:
- [x] E series should be a parameter, typically set by refinement design-wide, optionally overridden per-part
- [x] By default, it should not allow arbitrary resistances, but the user can override with a higher E-series or boolean flag

PR #65, opened 2021-10-29
Refactor resistive divider calculator

ducky64:
DO NOT MERGE - DEPENDS ON #59 

Previously, this used an algorithm specific to the resistive divider ratio and impedance equation, but would not generalize to other things - for example, non-inverting opamps or diffamps. This refactors the resistive divider calculator to use a generalized E-series ratio calculator, that walks through decades of values to search for a matching values and supports customizable criteria for determining additional decades to search..

Actually adding the additional parts will be a separate PR. Support for overriding single-resistor series (#16) will also be in a separate PR.

Additional changes:
- Fixes a tolerancing stackup bug in the actual ratio calculation - when terms appear twice in an equation it could cause larger-than-possible tolerances, this rewrites the divider formula to have terms only appear once.
- Renames the Exception module, since it was clobbering the actual Exception base class and causing weird things.


rohit507 (approved):
Seems ok.

PR #66, opened 2021-11-04
Refactor to use new Parts Table and remove old Product Table

### LINKED ISSUE #51: HDL interface server should start without parts tables
instead of just crashing and leaving us all very confused

### LINKED ISSUE #64: Refactor ProductTable -> PartTables
To use the new infrastructure from #59 

These additional classes need to be refactored:
- [x] Crystals
- [x] Diodes, Zener Diodes
- [x] FETs
- [x] MLCCs

ducky64:
Resolves #64. Resolves #51, since tables are lazy-initialized. One of those rare changes that has more deletions than insertions!

It was an interesting experiment in writing a product table DSL but it wasn't meant to be. As ugly as Python lambda syntax is, sticking to vanilla Python is probably the better choice.

This probably mostly preserves the prior behavior of equivalent parts. Some things have been refactored:
- Crystals now generate in the leaf type, and its parent can inspect its parameters such as to generate capacitors. Previously, the leaf type would be a dummy only.
- Parallel capacitors now generate into a dummy leaf type that just takes in parameters (previously, it used the same class which was a bit ehhhhh). The parallel capacitors algorithm has also been changed to account for total count and cost.
- Refactoring common Digikey column names into a shared base class.
- In general quite a bit of code cleanup and deduplication


rohit507 (approved):
Seems reasonable.

PR #67, opened 2021-11-05
Resistor refactor

### LINKED ISSUE #16: Resistor generator fixes

ducky64:
Resolves #16. It will no longer automatically choose a non-E-series resistor without explicit user action (set series to 0).

Refactors the E-series resistor classes to take a series and tolerance as parameters with a default value. These cannot (currently?) be set from within the design hierarchy (since not all resistors necessarily are E-series, for example if we have parts tables in the future), so would be set from refinements instead.

Also cleans up the preferred number chooser in the E-series util, with the rounding digits now hard-coded as a constant (it's really more a function of floating point precision and E-series precision, which I think tops out at 3 sig figs)

TODO
- [x] add constants for E192 series and E192+E24 series.


rohit507 (approved):
Looks reasonable.

PR #69, opened 2021-11-06
uploading JLC_Resistors.py

tengisd:


tengisd:
Adding Manufacturer part number

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
+class JLCPCB_ResistorTable(ESeriesResistor):
```
ducky64, line 1: I think you'll need some imports before this line

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+  RESISTANCE = PartsTableColumn(Range)  
+  POWER_DISSIPATION = PartsTableColumn(Range)  
+  PRICE = PartsTableColumn(float)
+  FOOTPRINT = PartsTableColumn(str)
+    
+  @classmethod
+  def _generate_table(cls) -> PartsTable:
+    RESISTOR_MATCHES = {
+    'resistance': "\s?(\d+(\.\d*)?[GMkmunp]?)?[\u03A9]\s",
```
ducky64, line 10: style: additional indent for dict contents. I generally follow IntelliJ's automatic styling, but with indents set to two spaces instead of the default four.

(also more indents needed below, but I won't flag every instance)

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+    'tolerance': "\s?([\u00B1]\d+(\.\d*)?%)\s",
+    'power_dissipation': "\s?([\u00B1]\d+(\.\d*)?ppm)/℃\s"
```
ducky64, line 12: Instead of using `\u00B1` and `\u03A9`, I think you can dump the unicode characters directly in the code?

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    def parse_row(row: PartsTableRow) -> Optional[Dict[PartsTableColumn, Any]]:
+      if (row['Library Type'] != 'Basic'):
```
ducky64, line 16: New thought: instead of hard-excluding extended parts, maybe just tack on the fee (I think $3 per component?) to the price column?

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
+class JLCPCB_ResistorTable(ESeriesResistor):
+  RESISTANCE = PartsTableColumn(Range)  
+  POWER_DISSIPATION = PartsTableColumn(Range)  
+  PRICE = PartsTableColumn(float)
```
ducky64, line 4: Consider creating a `JlcPcbTable` class that contains common definitions for all JLCPCB tables, like the `DigikeyTable` class in #66. Here, this would just be the price.

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def parse_row(row: PartsTableRow) -> Optional[Dict[PartsTableColumn, Any]]:
+      if (row['Library Type'] != 'Basic'):
+          return None            
+      if (row['First Category'] != 'Resistors'):
+          return None
+
+      new_rows: Dict[PartsTableColumn, Any] = {}
+      try:
+      # handle the footprint first since this is the most likely to filter
```
ducky64, line 23: style: contents of a try block should be indented

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+      footprint = ""
+      new_rows[cls.FOOTPRINT] = footprint
```
ducky64, line 25: I guess this is a TODO? Check out how footprint mapping is handled in the InductorTable?

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+    
+      extracted_values = parse(row['Description'], RESISTOR_MATCHES)
```
ducky64, line 27: I think I'd rather prefer the dict be dumped here instead of being declared separately. It kind of would be akin to a magic number, but I don't think the variable name helps us understand what this is doing more than a raw dict, and it also isn't being used anywhere else else.

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+    
+      new_rows[cls.RESISTANCE] = Range.from_tolerance(
+          PartsTableUtil.parse_value(extracted_values['resistance'][0], 'Ohm'),
+          PartsTableUtil.parse_tolerance(extracted_values['tolerance'][0])
+      )
+      new_rows[cls.POWER_DISSIPATION] = PartsTableUtil.parse_value(extracted_values['power_dissipation'][0])
+      new_rows[cls.PRICE] = float(max(re.findall(":(\d+\.\d*),", row['Price'])))
```
ducky64, line 34: I could see this being refactored into a JlcPcbTable class, since other parts will also use the same cost parser.

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+    
+      new_rows[cls.PRICE] = float(row['Price'])
```
ducky64, line 36: Why's this line here?

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+    
+      return new_rows
+      except (KeyError, PartsTableUtil.ParseError):
+      return None
+
+      raw_table = PartsTable.from_csv_files(PartsTableUtil.with_source_dir([
+      'JLCPCB_SMT_Parts_Library.csv'
+      ], 'resources'), encoding='utf-8-sig')
```
ducky64, line 44: I think you may want to change the encoding as we've discovered when trying to view the spreadsheet.

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+      new_rows[cls.POWER_DISSIPATION] = PartsTableUtil.parse_value(extracted_values['power_dissipation'][0])
+      new_rows[cls.PRICE] = float(max(re.findall(":(\d+\.\d*),", row['Price'])))
+    
+      new_rows[cls.PRICE] = float(row['Price'])
+    
+      return new_rows
+      except (KeyError, PartsTableUtil.ParseError):
+      return None
+
+      raw_table = PartsTable.from_csv_files(PartsTableUtil.with_source_dir([
```
ducky64, line 42: I think you need a de-indent here

electronics_lib/JLC_Resistor.py
```diff
@@ -0,0 +1,87 @@
...
+     self.assign(self.selected_resistance, part[JLCPCB_ResistorTable.RESISTANCE])
+     self.assign(self.selected_power_dissipation, part[JLCPCB_ResistorTable.POWER_DISSIPATION])
+
+     self.footprint(
+       'R', part[JLCPCB_ResistorTable.FOOTPRINT],
+       {
+         '1': self.a,
+         '2': self.b,
+       },
+       mfr=part['Manufacturer'], part=part['Manufacturer Part Number'],
```
ducky64, line 84: See the Inductor part refactor in #66, where the row names like `'Manufacturer'` are now a shared constant in the table base class.

ducky64 (commented):
The code looks fine for the overall structure, but I've made a few stylistic suggestions in the inline comments. Also needs unit tests, because I think the indents are wrong and this just won't work...

What's up with the PR being closed?

tengisd:
I closed the PR after realizing several mistakes in the code and also added
Manufacturer Part number into the code.

ducky64:
For unit tests: as a reference, check out the [inductor test case](https://github.com/BerkeleyHCI/PolymorphicBlocks/blob/master/electronics_lib/test_inductor.py). The expected values were created by running the code through and doing a quick spot check (looks about what I'd expect given the DUT setup).

ducky64:
For PRs: the typical development process is to continue pushing modifications / improvements to the same branch while keeping the PR open. The PR is more of the idea behind a code change, which can keep being updated until it gets merged.

Closing a PR generally means that something about it is fatally flawed - for example, the concept is wrong or perhaps the code needs a fundamental rewrite, instead of it just needs iteration.

tengisd:
Yes, I already wrote a unit test following it but got a trouble running it.
Would you be available for an online meeting tomorrow?

tengisd:
Oh, I see

ducky64:
Yep, tomorrow (today now?) works. We can schedule via Slack.

PR #71, opened 2021-11-09
Add precompiled JAR so things work out of the box

ducky64:
It will default to the development JAR if that's present though.


rohit507 (approved):
Good, this should help repeatability.

Issue #70, opened 2021-11-07
Adding more type information to exprs. 


rohit507:
As discussed on slack. My ideal way to add further type information would be by adding more ops/opcodes to expr.proto, so that we aren't overloading operations needlessly. 

In my ideal case (mocked up [here](https://gist.github.com/rohit507/d10ac6f98c921d6c3446052389e7e246)) we'd:
  - Add a unary expression that handles non-set unary operations. (lines 153-181)
    - Basic ops like NEGATE, NOT, and INVERT. (157-164)
    - Range value extraction ops like MIN, MAX, CENTER, and WIDTH (166-176)
  - Convert `ReductionExpr` to `SetExpr` (or similar) and use it for all unary operations with sets as input. (183-258) 
    - Convert MINIMUM and MAXIMUM to only apply to sets (217-227)
    - Add pointwise NEGATE and INVERT (246-254)
  - Remove unnecessary binary operations like SUB (43-44) and DIV (49-50), in favor of using the appropriate unary op to find the negation/inverse. 
  - Add BroadcastExpr for operations that apply an operation pointwise to all elements in a set. (138-151) 

This would require some minor changes to how things are encoded but reduced ambiguity and complexity a lot. For example: 
  - `a - b = a + (unaryNegate b)`
  - `1/sum(1/{a,b,c}) = unaryInvert( setSum( setInvert( {a,b,c} )))`
  - `{a,b,c}/3 = broadcastMult( {a,b,c} , unaryInvert(3))`
  - `3/{a,b,c} = broadcastMult( setInvert( {a,b,c} ), 3)`

Thoughts?

rohit507:
Updated the gist to make the various type signatures more explicit, also removed comments implying that subset and intersection are applicable to set-like types.

ducky64:
The mockup case looks reasonable, but some concerns:
- I can see how ADD / NEGATE and MULT / INVERT could be a more minimalistic encoding while being functionally equivalent. One question is how much we want to design the IR for usability - for example, should users (or rather, what kind of users) be expected to look at IR when debugging, The ADD / NEGATE would be a weirder encoding, though maybe the response to that is "users should only look at the Python code and the IR is a blackbox".
  - This also limits what tools can do with displaying expressions. But maybe the right answer there is instead of trying to pretty-ify IR, just find / link to the relevant line of source Python?
- Does it make sense to replace either LT or GTE, and GT or LTE with the NOT equivalent?
[...]

PR #72, opened 2021-11-09
Update IR to be more explicit by separating out ops on set-like values.

rohit507:
This has the updated expr.proto discussing in [this issue](https://github.com/BerkeleyHCI/PolymorphicBlocks/issues/70).

Broadly, this follows the changes describes there to separate operations for working with struct-like quantities.

There are a few differences from the version in that issue: 
- `ReductionExpr` was renamed to `UnarySetExpr` to better reflect that it contains both reduction and pointwise operations. 
- `BroadcastExpr` was renamed `BinarySetExpr` to better reflect that it *can* contain  binary set operations that don't just map over the set. For instance a true SUBSET op later on. 
- The definitions are reordered so that they go: `UnaryExpr`, `UnarySetExpr`, `BinaryExpr`, `BinarySetExpr` 
- The SUBSET binop has been renamed to WITHIN, which is functionally identical but the new name doesn't imply a set-like structure. Its semantics have not changed. 

Also, I've made sure the new protos compile but made no effort to update the python or scala code to compensate.

### DIFF
--- a/edg_core/Array.py
+++ b/edg_core/Array.py
@@ -5,9 +5,11 @@ from typing import *
 from . import edgir
 from .IdentityDict import IdentityDict
 from .Core import Refable, non_library
-from .ConstraintExpr import BoolExpr, ConstraintExpr, Binding, ReductionOpBinding, ReductionOp, FloatExpr, RangeExpr, \
+from .ConstraintExpr import NumericOp, BoolOp, EqOp, OrdOp, RangeSetOp, BoolExpr, ConstraintExpr, Binding, \
+  UnaryOpBinding, UnarySetOpBinding, BinaryOpBinding, BinarySetOpBinding, \
+  FloatExpr, RangeExpr, \
   ParamBinding, IntExpr, ParamVariableBinding, NumLikeExpr, RangeLike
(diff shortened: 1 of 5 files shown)


rohit507:
Okay, I'm not entirely sure how to handle `edg_core/Array.py`.

It seems like I ought to overload all the dunders with ones that properly use `BinarySetExpr` instead of just `BinaryExpr`, but that opens up the question of how to properly handle the case where there's an array on the rhs of the expression. Should the dunders for `ConstraintExpr` check whether there's an array on the rhs? maybe the best place to flip the args is when you're writing out the protos? 

Thoughts?

ducky64:
Dunders?

rohit507:
Double underscore functions like `__init__` or `__add__`.

ducky64 (commented):
The proto changes look reasonable. Will work through making the Python frontend and Scala compiler consistent with it.

For the proto: is the main distinction you're trying to make between variable-sized sets/arrays and scalars/fixed-sized sets (like ranges)? And it seems like it's mainly distinguishing by argument type instead of return type?

rohit507:
The extra commits on this request trying to push the changes forward are problematic, so #89 only has the proto changes.

PR #79, opened 2021-11-20
JlcResistors

tengisd:


tengisd:
Codes only related to JLC resistors are pushed by this branch

electronics_lib/JLcTable.py
```diff
@@ -24,11 +23,11 @@ def _parse_jlcpcb_common(cls, row: PartsTableRow) -> Dict[PartsTableColumn, Any]
       cls.COST: float(max(float_array)),
     }
 
-  """Extracts out component values from the "Description" column of JLCPCB_SMT_Parts_Library.csv"""
+  """Extracts out component values from the "Description" column of JlCPCB_SMT_Parts_Library.csv"""
```
ducky64, line 26: Typically docstrings are just one `"""` block, instead of split between multiple blocks. Examples can go into the same docstring as the main description.

electronics_lib/JLcTable.py
```diff
@@ -1,18 +1,17 @@
 import re
 from .PartsTable import *
 import numpy as np
```
ducky64, line 3: Can you get rid of the np import?

electronics_lib/test_JLcResistor.py
```diff
@@ -1,22 +1,21 @@
 import unittest
```
ducky64, line 1: filename should be updated to be consistent

electronics_lib/test_JLcResistor.py
```diff
@@ -1,22 +1,21 @@
...
-from .JLcResistor import JLcResistor, JLcResistorTable
+from .JlcResistor import JlcResistor, JlcResistorTable
```
ducky64, line 2: Tests are failing - I think you renamed the JlcResistor class but not the file.

For reference: `from .JlcResistor` looks for a file named `JlcResistor.py`, and it's case sensitive.

ducky64 (changes requested):
A few remaining issues, but looks better.

Can you make sure mypy is clean and the test pass? We should really set up GitHub Actions CI at some point, but that point sadly isn't right now...

electronics_lib/JlcTable.py
```diff
@@ -1,10 +1,9 @@
 import re
 from .PartsTable import *
-import numpy as np
-class JLcTable(LazyTable):
-  """Shared base class for JLCPCB product tables that contains common row definitions."""
 
-  DISTRIBUTER_PART_NUMBER = 'LCSC PART'
+class JlcTable(LazyTable):
+  """Shared base class for JlCPCB product tables that contains common row definitions."""
+  DISTRIBUTER_PART_NUMBER = PartsTableColumn(str)
```
ducky64, line 6:
```suggestion
  JLC_PART_NUMBER = PartsTableColumn(str)
```

This is specific to JLC - the only use for this is to generate JLC pick'n'place data. Unlike manufacturer part number, this (probably) isn't going to have any more general usage, so it's safe to name it JLC-specific.

(and also change other references for consistency)

electronics_lib/JlcResistor.py
```diff
@@ -66,24 +68,24 @@ def __init__(self, **kwargs):
...
+      'R', part[JlcResistorTable.FOOTPRINT],
       {
         '1': self.a,
         '2': self.b,
       },
-      mfr=part[JLcResistorTable.MANUFACTURER], part=part[JLcResistorTable.PART_NUMBER],
-      value=f"{part[JLcTable.DESCRIPTION]}",
-      datasheet=part[JLcResistorTable.DATASHEETS]
+      mfr=part[JlcResistorTable.MANUFACTURER], part=part[JlcResistorTable.PART_NUMBER],
+      value=f"{part[JlcTable.DESCRIPTION]}",
```
ducky64, line 89:
```suggestion
      value=part[JlcTable.DESCRIPTION],
```

Don't need the f-string / string interpolation if it's just one variable

ducky64 (commented):
Almost, just two more nits - hopefully the last lteration.

ducky64 (approved):
lgtm!

PR #82, opened 2021-11-20
Automated PR checks

ducky64:
- [ ] Python type checking
- [ ] Python unit tests
- [ ] Scala compile
- [ ] Scala unit tests


ducky64:
@rohit507 I'm merging this now because it's pretty useful but do take a look at the workflow code post-merge

@tengisd this fixes a case-sensitivity issue in one of your tests - the tests probably ran fine on Windows because Windows is case-insensitive, but it fails on the GitHub Linux environment.

PR #89, opened 2021-12-20
Update IR with more explicit set operations. 

rohit507:
This is a version of #72 with the extraneous non-proto commits removed.


ducky64:
Obsoleted by #90

PR #90, opened 2021-12-21
PR 72/89 consistency updates

ducky64:
Refactors code to work with the proto changes from PR72/89.
- Most of the Python frontend code was fixed in PR72/89, but there were a few stragglers fixed here

Unfortunately this pulls in several other refactors, which probably accounts for the bulk of the non-generated-code diff:
- Restructure the proto so that mypy-protobuf can automatically generate package names, instead of relying on the goshawful sed script
- Associated renaming in Python and Scala imports

Includes commits from PR #89


ducky64:
All tests passing, ready for review.

rohit507 (approved):

Issue #95, opened 2022-02-04
IR cleanups


ducky64:
After single-shot generators is consistent, these changes are on the pending list to clean up the compiler:
- Separate expressions (things that evaluate to a value, eg binary operators) from statements (top-level constraints that have no value, eg connects, assigns, asserts (new))
- Add error types in the structural IR (eg, BlockLike can also be a BlockError message, or even just Error message), which would also add error handling into the interchange format (instead of being side-channel data).
  - This would only be for errors that prevent the structural IR from elaborating. Assertion failures would be handled separately - either continuing to be side-channel data, or with some other construct (eg, error / assertion failed statement type).

PR #75, opened 2021-11-13
Compiler refactor

lelzeiny:
Refactored so that allocate ports are lowered into individual ports prior to processing their constraints.

Issue #33, opened 2021-02-27
Port model constructor refactoring: eliminate passing port model, support required parameters


ducky64:
For example, `DigitalBidir` can take a port model (another DigitalBidir) as a constructor argument, essentially cloning it. This is a confusing pattern as it requires support in each port definition, and presents an inconsistent interface where it isn't supported.

The current use case seems to be in Bundle definitions, like in the UART where
```python
    self.tx = self.Port(DigitalSource(model))
    self.rx = self.Port(DigitalSink(model))
```

It's "clean" in that it lexically has the `self.Port(DigitalSource(...))` structure, and supports model=None, but "unclean" in that it's passing a port model into the constructor for another port model, which is not a general pattern.

There are other use cases, such as extracting source-only / sink-only parameters from a DigitalBidir. Alternatives may be DigitalSource.fromModel or DigitalSource.fromModelOptional.

This bug only applies to cloning / casting use cases (is-a relationship), and does not apply to composition use cases (has-a relationship, eg passing a port model into a bundle, like the UART example above)

ducky64:
Related: add support for required parameters for port construction.

This would still need a null-parameter set for port construction, for link-side instantiation and library construction. Maybe `Port.empty()`?

ducky64:
A bit of a braindump going over possible refactoring:

Two main directions: removing port models to be more explicit (eg, `DigitalSource.fromBidir(...)` - pretty straightforward) and rethinking the default empty port (what does `DigitalSource()` mean? - needs more thought).

For (as an example) `DigitalSource()`, there are currently a few use cases:
- Declaring it as a link port, which needs to have uninitialized parameters. This can be more verbose (eg, `self.Port(DigitalSource.empty())`) since this is a power user feature. I believe current semantics have links discard port initializers, but instead this should be check-and-error or check-and-warn.
- Declaring a base class port. As the class shouldn't be in a final design (and causes an abstract class error), the semantics here are less important.
  - In any case, it needs to be override-able by a subclass. Current behavior has it creating a port with the default ideal initializer, but that gets dropped when it is over-assigned. Convenient semantics, but "magical"
  - One option would be to initialize it an empty (implicitly or explicitly). This will require all subclasses override all parameters.
[...]

Issue #26, opened 2021-02-15
Consider stricter over-assign semantics


ducky64:
With the new compiler (refactor in progress), equality propagation when both values are equal is silently discarded. This is needed to support the DC-DC converter case, where the voltage is forward-declared from the feedback resistor, before the generator runs and hooks up the inductor which should define the output voltage.

Potential alternatives:
- Explicit forward-declaration tag / annotation to enable this discarding (and erroring otherwise)
- Allow merging of nets between generator functions (might still require other funkiness in the DC-DC converter code)
- Other DC-DC converter restructuring to avoid this behavior

ducky64:
Probably resolved with directed assigns and DC-DC re-structuring.

PR #96, opened 2022-02-05
Single-shot generators and compiler refactor

### LINKED ISSUE #28: Unified and coherent separation of specification and selected parameters
... and how specification component can be ignored when a subclass is set.

Use case:
- specifications are hack-ignored: Battery subtypes, PPTC fuse subtypes
- things could be cleaner in: all the part-table-based components, particularly resistors / caps / inductors.

### LINKED ISSUE #29: Allow split nets in generators
Currently, links must be fully defined in a generate function or in the base block definition - they can't be split. This makes for some awkward code (though generators are advanced functionality).

One idea to fix this would be to defer link size resolution until all the generators that could attach to a link runs (perhaps sharing infrastructure with #20). Generators would need to declare where they connect, though.

Alternatively, generators could define the block's implementation so the block doesn't elaborate until its generator runs. Which, if it doesn't break any use cases, could be the more elegant solution. This limits to one generator per block.

### LINKED ISSUE #63: Take another look at and probably refactor the compiler structure
The compiler structure could use another look as feature creep have pushed it beyond the initial clean-ish design. A refactor may be in order...

## Current compiler flow & requirements
Overview: the compiler runs until the outstanding work dependency graph is empty. Each unit of work is an ElaborateRecord - which generally elaborates (expands out) some aspect of the design. The graph starts with an unelaborated top-level block that is ready to go (no dependencies).

The compiler also maintains a ConstProp data structure, which handles parameter assignment and resolves parameter values. It is given constraints (assignments, forced-assignments such as from refinements, bidirectional equality such as from connects), and automatically evaluates expressions as dependencies are resolved. Assertions are stored separately in the top-level compiler.
(TODO: can we get rid of the bidirectional equality special case?)
(TODO: is there a cleaner way to support class-based parameter refinements?)

These supplemental data structures also exist (TODO can we get rid of them / unify them with something else?)
- `connectedLinkParams`: port to its connected link and all params in that link (TBD: recursive?)
- `directConnectedPorts`: all ports that have a direct connection (external side)

These are the elaboration operations:
- **Elaborate block** (entry point)
  - Check for refinements (by path, then by class), and instantiate block
  - Add class-based value refinements (TODO is this the right place for this?)
  - Instantiate ports in-place (replace with library elements? TBD more details)
  - Process constraints
    - `processBlocklikeConstraint` (needs a better name): handles assignments and assertions
    - collect all ports involved in link arrays - to size link ports
      - TODO: the direct connect case is expanded here, perhaps it should be delayed to unify with the array connect case?
      - TODO: the over-connect checks can probably happen here, but need some idea of hierarchy and levels of connection
  - For each connection, generate a ElaborateRecord.Connect, dependent on the port's block, the link, and when CONNECTED_LINK is known [TBD: ???]
    - Exports are dependent on the port's block, the current block (effectively a nop?), and when CONNECTED_LINK is known
  - For any generators: generate an ElaborateRecord.Generator, dependent on required parameters and ports
  - For any internal blocks: generate an ElaborateRecord.Block, dependent on own generators (since internal blocks need to know connected state on ports) and external block connected state
    - TODO: can these conditions be unified into a BlockFullyElaborated item (instead of dealing with them piecemeal), and sub-blocks can't start elaborating until its parent is complete?
  - Link this newly instantiated block into the containing (parent) block
- **Elaborate generator**
  - Dependency parameters values are collected
    - Currently, this also includes (implicitly) IS_CONNECTED and CONNECTED_LINK.NAME for edge ports. TODO: can we make these dependencies explicit, and how can we support CONNECTED_LINK.NAME in the case where IS_CONNECTED is optional?
  - Run the generator, passing in the parameter values
  - Merge the result into original block
- **Elaborate links** (quite similar to block case)
  - Set name variable (TODO this should be done in the other cases too)
  - Process constraints
    - same `processBlocklikeConstraint` for parameters
    - collect all ports involved in arrays (Export with MapExtract)
      - TODO the direct export case is expanded here, perhaps delay to unify?
  - Expand inner links, including arrays
    - For each leaf connection, generate a ElaborateRecord.Connect, dependent on the self link, the inner link, and that its CONNECTED_LINK is known [TBD: ???]
    - And add the direct connected ports of the inner port
    - Generate the new leaf constraint for the exported constraint (TODO: split the expand and elaborate phases?)
    - Set array elements and length parameters of inner port arrays
- **Elaborate connect**
  - Propagate (generate equality for ConstProp) for parameters between the connected ports, as well as CONNECTED_LINK sub-parameters
  - Adds ElaborateConnect records for exports for inner links, dependent on when the links resolve
  - TODO: instead of bidirectional equality constraints, can we instead have fully unidirectional constraints, mainly that parameters propagate towards the link? So the link can't assign any port parameters (except through CONNECTED_LINK)?
- **Elaborate block ports connected** (for block-side ports)
  - For non-connected ports, set just the name (but parameters otherwise empty)
  - Recursively set is-connected status
- TODO: where should things be mutable? Can we contain mutable effects, or make the individual elaborate operations more functional?

Summary of requirements
- Instantiate blocks (and its ports and links, recursively)
- Handle parameters evaluation, both from assignment constraints and connected equality
- Handle generators, dependent on required parameters and ports
- Handle IS_CONNECTED, especially the false case which can only be known after the block is fully elaborated (generators run)
- Handle values of special parameters IS_CONNECTED and NAME
  - IS_CONNECTED only used in MCU pin assignment, as well as an auto-generated constraint in non-optional ports
  - NAME also only used in MCU pin assignment (? - find usages isn't great because `name` gets used in a bunch of other contexts too)
  - Note: IS_CONNECTED does NOT imply that CONNECTED_LINK exists, IS_CONNECTED is a local operation only. Should this be the case? Can we require that IS_CONNECTED implies CONNECTED_LINK exists, even if it only connects to a dummy block?
- Handle link arrays, including nested link arrays

## Relevant issues
- #48 should probably be implemented in the compiler flow - so failures can be piped through and highlighted in the IDE instead of failing to compile completely
- #20 almost certainly requires a significant addition to the compiler, possibly even structural changes to support it

### LINKED ISSUE #92: Single-shot generators
Perhaps not the most accurate title.

Implementation path to address #29 by preventing the feature - generators would no longer append nets / connections, rather the single generator on a block would define all the connections. This may also simplify parts of the compiler logic, since there was logic and state to handle partially defined blocks.

Before the generator runs, these would be defined on the block:
- Parameters
- Parameter assignments
- No internal blocks are defined
- No connections are defined

The generator then defines these parts of the block:
- Internal blocks
- Connections (between internal blocks, and between boundary ports and internal blocks)
- ... including any parameters

Example issues: buck converter generator
- Actual output voltage is dependent on the feedback resistive divider - and dependent on internal blocks
- The inductor sizing is dependent on the current draw
- There are some components where the current draw is dependent on the voltage (perhaps this is bad practice and can lead to brittle design? perhaps listed voltages and currents must be independent of each other to avoid deadlock?)
- Possible solution: move the generator boundary for the buck converter to internal blocks. For example, instead of the buck converter block generating the power stage, the power stage would be a separate generator block contained within a non-generator buck converter block. This allows the feedback resistive divider and power stage to generate independently, but at the cost of more levels of hierarchy / blocks. However, this structure might enable better re-use in the power converter generator through better modularization.
  - In general, the idea would be to have very minimal generators (instead of more monolithic generators). Downside is that it may mean more individual sub-blocks so we can have those generator boundaries. This structure (top-level block is just a non-generator encapsulating block) also avoids needing the generator to propagate internal block information that the generator needs to run.

Alternatively, a pre-generator block can be more akin to a library block, where it defines nothing at all. This would probably simplify and unify the block elaboration logic.

ducky64:
Major compiler refactor and small semantics overhaul:
- Resolves #92, constrain generators to single-shot semantics
  - Resolves #29 by obsoletion
  - Runtime check only init args are in generators
- Resolves #63, general cleanup of the compiler code
- Preparation for #20, block-side port arrays
- Constructor arguments (with `@init_in_parent`) create parameters by default, and these should not be redefined in HDL
  - Constructor arguments will infer parameter types based on (now required) type annotations.
  - Defaults no longer needed on constructor arguments - so the type system can check for missing arguments in block instantiations.
- Rename footprint parameters with a `fp_` prefix to avoid aliasing with user constructor arguments in subclasses
- Resolves #28, by introducing a convention: parameters are spec by default, while actual values are prefixed with `actual_`. This only applies for blocks, Ports and Links always model device limits (for `_limit` parameters) or actual values. Syntactic sugar can come in a future PR, but this effectively solves the problem.
- Update all the examples for the new syntax / semantics.

Minor cleanups:
- Refactor the link array expansion to the lower-to-individual-connects style for blocks introduced in #75
- Style: unify Scala unit test compile / error checks
- Style: ArgParameter or something to get rid of the nasty casts (sub-PR #101)
- Style: eliminate contents() for simple wrappers (eg, resistor types) (sub-PR #101)

These related items will be done in a future cleanup PR:
- #95, propagating the single shot generator semantics through to the IR
- #33 and #12, as a extension of the constructor argument cleanup here
- #26, now that the DC-DC converters have better semantics

These things may need more thought in the future:
- IS_CONNECTED is set in a few places: in the connect statement forwarding the value (bidirectional equivalence), in the link port elaboration (set IS_CONNECTED to true), and in the block port elaboration (in the not-connected case, if there wasn't a connection in the parent block). The last one is the ugliest (though not horrifyingly so), but is needed in that port must be known to know which ports are not connected.
  - Potential alternative: require explicit non-connected statement in the frontend (this could be a dummy block). IS_CONNECTED could instead be a link parameter instead of a special case. This is already needed in some cases.


rohit507 (approved):
Looks good to me.

PR #100, opened 2022-02-12
Support default parameters

ducky64:
Note that this merges into another development branch, it's a required feature for the simplified generator semantics.

This adds support for default parameters values in blocks in the proto IR. Default parameter values are (in the HDL) constructor parameters that are defined with a default value, eg `def __init__(param: IntLike, default_param: IntLike = 42)` would have a default for `default_param` of `42`. When the compiler encounters a refinement, it looks for default values for parameters in the refined block that don't exist in the unrefined block, and sets the parameter value. This should be safe, since the design would expect the unrefined type and wouldn't know about (and be able to assign) parameters that only exist in the refined type. Parameter refinements coexist with this and take precedence over default refinements.

Example use case: the default resistor class only defines resistance, but the chip resistor also specifies tolerance and footprint. To allow the refinement (and the generator to evaluate), the chip resistor sets a default tolerance and footprint. Since it's all in the constructor, it's consistent whether it's set by refinement or directly instantiated (in which case the defaults are initialized in the parent block).

### DIFF
--- a/edg_core/test_default.py
+++ b/edg_core/test_default.py
@@ -11,7 +11,7 @@ class BaseParamClass(Block):  # Base class containing no parameters or default
 
 class NondefaultParamClass(BaseParamClass):  # contains a single non-default param
   @init_in_parent
-  def __init__(self, nondefault_param: IntLike) -> None:
+  def __init__(self, nondefault_param: IntLike = IntExpr()) -> None:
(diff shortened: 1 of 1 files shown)


ducky64:
@rohit507 Merging as-is since this block development, but feel free to review and make suggestions.

Issue #98, opened 2022-02-10
More accurate constructor argument typing


ducky64:
Currently, constructors (`__init__`) take `*Like` types (eg, `StringLike = Union[str, StringExpr]`), which was needed to allow literal value shorthand in constructor calls. However, the `@init_in_parent` provides a translation later that actually narrows the types, `__init__` still accepts the `*Like` types on a call, but internally (seen by `__init__`) is all `*Expr` types.

Prior to #96, all constructor parameters should be re-mapped via `self.param = self.Parameter(...)` and given an explicit type. #96 changes the constructors to infer parameter types based on typing annotations, so for example if we had `param` as part of the constructor argument list, we could directly use `self.param = param`. The downside is that `param` is of type `*Like` (from the constructor) instead of the more narrow and accurate `*Expr` (returned by `self.Parameter`).

This causes a mild issue with generators, since they should only take `*Expr` types. The current workaround is to expand generators to also take the castable-types in addition to the `*Expr` types - which isn't completely accurate, but continues to provides convenient lightweight syntax in exchange for a bit of leakiness in the type checking of an advanced feature.

## Properly typing `init_in_parent`
Ideally, one solution would be to define the `init_in_parent` decorator type, such that the `__init__` can be properly defined in terms of `*Expr` type, and the decorator would provide a wrapped function that takes the `*Like` type. The problem is that functions (Callables) have contravariant types for the argument list, so it can't capture the ConstraintExpr[...] type parameters (for example, it doesn't accept BoolExpr in place of the ConstraintExpr[...] pattern).

Additionally, having an *Expr in the type signature of `__init__`, even if the decorator is properly typed, may be confusing, since users might see the `__init__` type signature and wonder why a literal type is allowed. However, the `@init_in_parent` decorator could signpost that, and since the behavior is consistent it's only one additional learning curve item.

PR #101, opened 2022-02-16
Use self.ArgParameter for __init__ argument list parameters

ducky64:
Not the perfect solution, but cleaner than `cast(RangeExpr, initParam)`. This is the current workaround for #98.

Also removes the implicit promotion from float to Range, which while convenient, really probably isn't safe. Still allows range-float operations, eg range minus float.

Also a few additional cleanups in the standard library.

### DIFF
--- a/blinky_skeleton.py
+++ b/blinky_skeleton.py
@@ -1,7 +1,7 @@
 from edg import *
 
 
-class BlinkyExample(SimpleBoardTop):
+class BlinkyExample(BoardTop):
   def contents(self) -> None:
     super().contents()
     # your implementation here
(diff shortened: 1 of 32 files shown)


rohit507 (approved):
Looks good to me.

I'm curious, are all the `tol` parameters on the units mandatory? is the old syntax `1*Volt` still valid or is `Volt` a function now, rather than a callable?

ducky64:
The `tol` parameter is necessary for a float -> range conversion. `1 * Volt` (I think) still produces a float-like, the difference now is that there's no implicit promotion (in the static typer, at least) from float to range.

ducky64:
The underlying reason (other than 'it feels magical and maybe a bit dangerous') is that the float-like and range-like types overlap, so this change of removing implicit promotions removes that overlap.

PR #102, opened 2022-02-28
Eliminate port model argument

### LINKED ISSUE #33: Port model constructor refactoring: eliminate passing port model, support required parameters

ducky64:
Resolves #33

Previously, DigitalSource and DigitalSink had a first `model` argument which allows them to initialize from a DigitalBidir (or another DigitalSource or DigitalSink). This removes that argument in favor of an explicit cast, `DigitalSource.from_bidir(model)`.

All examples are fixed to be consistent.

Other changes:
- This also fixes the behavior of `get_enclosing_block` for scope checks - checking the enclosing block but allowing it to be in a different (eg, port initializer) context.
- NotConnectablePort also cleaned up slightly, but the actual functional change was backed out as it wasn't useful.
- Fixes PortBridge to generate inner blocks

### DIFF
--- a/edg_core/Builder.py
+++ b/edg_core/Builder.py
@@ -20,11 +20,13 @@ class Builder:
     self.stack.pop()
     assert self.get_curr_context() is elt
 
-  def get_curr_block(self) -> BaseBlock:
+  def get_enclosing_block(self) -> Optional[BaseBlock]:
     from .Blocks import BaseBlock
-    elt = self.stack[-1]
-    assert isinstance(elt, BaseBlock)
-    return elt
+    # traverse down the stack and get the first BaseBlock, ignoring things like ports
+    for elt in reversed(self.stack):
+      if isinstance(elt, BaseBlock):
+        return elt
+    return None
 
   def get_curr_context(self) -> Optional[Refable]:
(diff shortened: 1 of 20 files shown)


rohit507 (commented):
LGTM

Issue #20, opened 2020-12-24
Block-side Port Arrays


ducky64:
Potential implementation
- LinkArray construct, that allows an arbitrary-length parallel array of links. It has ports that remap to the internal links.
  For example, if the internal link is a DigitalLink with `src: DigitalSource` and `sinks: Array[DigitalSink]` ports, the LinkArray would have:
  - superclass `DigitalLink`
  - (once fully elaborated) ports `src: Array[DigitalSource]` and `sinks: Array[Array[DigitalSink]]`
  - (once fully elaborated) constraints to connect `array.src[x] <=> array.links[x].src` and `array.sinks[0][x] <=> array.links[x].sinks[0]`
  - Its length is equal to the length of all of its ports (bidirectional propagation / merged construct). It is ready to elaborate when any port's length is known.
  - **TODO**: how should width inference be handled in the constraint language - or should it be a special construct? Can it work for a fully directioned constraint language - in this case, where are the constraint directions set?
- `LENGTH` ReservedParam, which can be used on a PortArray to get a Int type corresponding to the length.
  - This is the mechanism for width inference. This `LENGTH` parameter can both be a value source (eg, in a generator, get the actual length of the block's ports) and a sink (eg, set the length from a generator)
- `ALLOCATE` ReservedParam, which can be used on a block-side PortArray to get a fresh "slice" of ports of the array. Dynamic length (length propagates to the block), and can be connected to either an element port (single link connect) or array port (link array "parallel" connect). The PortArray's length is the sum of the length of all ALLOCATE operations. Any PortArray can either be connected as a whole, or piecewise with its ALLOCATEd components (but never both - mutually exclusive).
  - The motivating use case is a microcontroller's digital IOs, which might consist of a mix of known-width connects and unknown-width (at compile time, but resolved by parameter propagation width inference in the backend) connects, and to make that representable in the IR.

### Alternatives
- Instead of an PortArray.ALLOCATE operation, perhaps something can be done with a make_array expression, eg `connect(make_array(io0, io1), link.io_array)`. But this doesn't capture the microcontroller digital IOs use case well, and would also collapse the links from multiple connect statements into one mega link-array which could make visualization and debugging less intuitive.
- However, make_array might be a good for for link-side array allocation - instead of connecting block ports to link.ios, all the block-side IOs in the connection can be grouped into one connect statement. Potentially simplifies the Python fake-backend code.
[...]

PR #105, opened 2022-03-14
Block-side port arrays

ducky64:
Partial resolution for #20, adds block-side port arrays with:
- Ability to define and export port arrays in blocks in the frontend
- Block side port arrays can have undefined elements (useful for an abstract block, is an error if this shows up in the final design), or defined elements.
- Port array elements can be defined statically as part of a block, or 'dynamically' (including reactive to incoming connections) as part of a generator. Array elements can also be defined by propagation (exports).

A future PR will complete #20 with these features:
- Port array to port array connections (between blocks at the same level, as opposed to cross-hierarchy export operations)

A work-in-progress PR will use this feature to add an abstract microcontroller block (#23) and refactor examples to use that. That may also modify the frontend interface for port arrays.
- TODO: get rid of required-ports, which is hackish and used by the current microcontroller generator.

---

Implementation details:
- IR changes
  - Add separate array-export and array-connect constraints, to differentiate from the single element export and connect constraints
  - For Port Arrays, differentiate between undefined and empty with a contains element.
  - Add Array-valued ValueExpr
  - Change ALLOCATE reserved-param (LocalStep) to an allocate step with optional suggested name
  - Add ALLOCATED reserved-param (LocalStep) which returns the requested connections to a port-array as an array-of-string
    - ALLOCATED propagates root-inwards, and is defined by the top-most-level connections
  - Add ELEMENTS reserved-param (LocalStep), which returns the elements of a port-array as an array-of-string
    - ELEMENTS propagates leaf-outwards, and is defined by statically defined ports of an array
    - This is more of an internal thing and is not exposed in the frontend
  - Minor refactorings
    - Clean up generator to get rid of function name and other legacy cruft. (as part of #95)

- Major refactoring of compiler to make this all work
  - Block and links now expand as part of their container. This elaborates their ports (so IsConnected can be known).
    - Exception: port arrays have delayed elaboration, until ELEMENTS is defined (which can be dependent on ALLOCATED).
    - All link-side port arrays are treated as dynamically sized by incoming connections, so this defines a directed ALLOCATED -> ELEMENTS assignment.
    - Their contents are processed separately
  - Per the above, the generator task is integrated with the rest of the block task. The generator's block stub includes the port definition (excluding arrays), and the implementation includes the internal blocks / links / constraints.
  - IsConnected handling is now (more) centralized, enabled by the above change.
  - Add separate lowering of array-connect constraints (expands into leaf-level connect constraints - and defines ALLOCATED) and ALLOCATE references (which assigns them a concrete index)
  - Additional stylistic changes
    - Add a single-write hash map data structure that builds in check for overwriting or multiple writes
  - Add compiler error check for undefined port-arrays (such as might be seen on an abstract block, distinct from empty port arrays)
  - Add compiler check pass for bad references to ports in connects and exports.
    - These do not check parameter references, since connectivity to the link is needed to resolve steps like CONNECTED_LINK.

- Frontend changes
  - Add frontend support for Port Arrays: append_elt (from the inside) and allocate(suggestedName) (from the outside)
  - Add Vector.allocated() as a way to get incoming connections (including suggested-names) as a string-array
    - Add generator support for array-valued ValueExpr
  - Minor refactorings
    - Eliminate intermediary ParamVariableBinding that wasn't doing anything useful
    - Split out GeneratorBlock into its own file


ducky64:
Not sure what's going on with automated tests, it seems to think Builder.get_curr_block() doesn't exist?

rohit507 (approved):
lgtm

PR #106, opened 2022-04-01
Port.empty() refactor

### LINKED ISSUE #103: Define default semantics for port instantiations
Supersedes #12, extension of #33

Currently, a port with no arguments (eg, `DigitalBidir()`) means a few things in different contexts:
- In a block where it is the leaf port (not an intermediate / exported port), it means to create a port with ideal behavior and parameters (eg, infinite voltage limits, empty output voltage)
  - Open question: Should ports have required parameters? What (if any, or how many) defaults should be provided? Should ideal ports be explicit (eg, `DigitalBidir.ideal()`)? **This is the main open question**
- In an block where it is an exported port (whether with `self.Export(...)` or `self.Port(...)` that is later connected), its values are discarded. Ideally, it would be checked that the user did not explicitly specify values, but this isn't implemented.
  - In most cases, `self.Export(...)` is preferred. However, one issue here is that it prevents the `__init__` and `contents` split that provides runtime efficiency. **This is the main problem case**
  - One case where `self.Export(...)` isn't possible is abstract base classes. Since this is a more advanced feature, requiring an explicit empty (eg, `DigitalBidir.empty()`) is probably acceptable.
  - One benefit of an explicit `empty()` is that a subclass could override it later. Perhaps `init_from` should come back, allowing port initializers to be overridden as a unit rather than piecemeal with parameters.
- In a link, it means to create an empty port (containing no parameter value - not just ideal parameter values). On links, port parameters are (by convention) always taken from the block-side ports. There is some (inconsistent) refactoring where this is made more explicit, eg `DigitalBidir.empty()`.
  - Similarly, as this is a more advanced feature, requiring an explicit empty is probably acceptable.

ducky64:
Resolves #103 , forcing Port.empty() to be used for intermediate ports, and refactors existing code to push this through. Defines Port.empty() in infrastructure (structured as clearing all initializers)

Also:
- Refactor the port initializer code to be centralized in Port, in preparation for IO arrays.
- Clean up a bunch of cruft
- Get rid of the ignore_* in the creation of proto - this silently drops data
- Internally, initializers are treated as mutable (in infrastructure only) and copied on a clone operation. Previously, all state must have been captured in the __init__ arguments, which is too limiting for heterogeneous IO (eg, a SPI port consisting of micro pins with different digital characteristics). This enables that.
- Refactor the Ground() not-a-port port so it can also support Ground.empty()
- Refactor this through all the examples and libraries

### DIFF
--- a/edg_core/Array.py
+++ b/edg_core/Array.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 from typing import *
 
 import edgir
@@ -146,6 +147,9 @@ class DerivedVector(BaseVector, Generic[VectorType]):
     assert base.parent is not None  # to satisfy type checker, though kind of duplicates _is_bound
     self._bind_in_place(base.parent)
 
+  def _type_of(self) -> Hashable:
+    return (self.target._type_of(),)
+
   def _get_elt_sample(self) -> BasePort:
     return self.target
 
@@ -155,8 +159,8 @@ class DerivedVector(BaseVector, Generic[VectorType]):
   def _def_to_proto(self) -> edgir.PortTypes:
     raise RuntimeError()  # this doesn't generate into a library element
 
-  def _type_of(self) -> Hashable:
-    return (self.target._type_of(),)
+  def _get_initializers(self, path_prefix: List[str]) -> List[Tuple[ConstraintExpr, List[str], ConstraintExpr]]:
+    raise RuntimeError()  # should never happen
 
 
 # An 'elastic' array of ports type, with unspecified length at declaration time, and length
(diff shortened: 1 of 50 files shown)


ducky64:
This is blocking the array IO refactor and is frontend only, so I'm just going ahead and merging this. @rohit507 feel free to do a post-merge audit if you'd like.

I'm not completely happy with the syntax: I think eg. DigitalBidir() should be empty, and we should have DigitalBidir.ideal() or something for the ideal case, not the other way around as currently implemented, but this is difficult in Python. But because the semantics are split now, static analysis tooling can help us change that if desired in the future.

Issue #97, opened 2022-02-05
Multimeter examples TODOs


ducky64:
Definitely was a rushed design.

- [x] Ranging circuit: divider needs to go to the input negative, not GND (blue wire fix confirmed in testing)
- [x] ~Ground negative input, use a switched cap inverter + diff-amp to re-center negative voltages~ may not be a good idea
  - Note: scope node used LM2660
  - Inverter selection:
  - Dual-rail analog switch selection
    - SPST switches: DG9421/2, ADG601/2, MAX4675/6 are pin-compatible
      - DG467/8 and DG447/8 are compatible and inexpensive
    - Note: 2:1/1:2 muxes are uncommon / unavailable for dual-rail
- [x] ~Actual clamping diodes~ the 1M input resistor is probably good enough, if it's high volt rated
- [ ] High voltage input protection - resistor, fuse (PPTC or otherwise), TVS suppressor (Zener diode or MOV) - suppress input voltage down to levels within the device ratings.
- [x] nRF: VBUS needed if USB is being used (blue wire fix confirmed in testing)
- [ ] Ranging circuit: don't need the 1:10000 step (at 4096 counts, centered around zero, this gives a 30kV full scale (at like 4 volts per bit), which will fry everything else)
- [ ] Ranging circuit: support more ranges, depends on #20, probably a 1:10, 1:100 step, maybe also a 10:1 amplification step
  - 10:1 amplification unnecessary, MCP3561 has internal amplifier 1/3 to 64x
- [x] Speaker circuit: needs to be redone, ~the fake-ADC does not work well for audio-frequency signals (could also be a software issue)~ (it was a software issue, mbed's PwmOut is limited to 1us increments)
  - [x] Use a class-D amp for efficiency. TPA2005D1 and PAM8302A are pin-compatible in MSOP-8. Use single-ended mode.
  - [x] Lower the filter cutoff frequency to like 2kHz
- [x] Speaker to use a base Connector class
- [x] Add battery measurement circuit (optionally w/ disable?)
- [x] nRF: perhaps replace with easier to source (but also difficult-to-solder LGA instead of castellated pads) MDBT50Q-1MV2
  - Note: MDBT50Q-512K (nRF52833) also exists and appears partially pin compatible. No mbed support yet.
- [x] ~Transflective LCD~
[...]

PR #107, opened 2022-04-04
Abstract microcontroller type

### LINKED ISSUE #23: Abstract microcontroller type

### LINKED ISSUE #30: Eliminate self.get, self._has for generators
The generator functions receive the unpacked parameters, but a way is still needed to replace `self.get(some_port.link().name)` and `self.get(some_port.is_connected())` without like a 100-long argument list for microcontroller pin assignment.

ducky64:
Resolves #23, by adding an abstract IoController type that provides dynamically-sized arrays of common IOs: GPIO, ADC, DAC, I2C, SPI, UART, CAN, USB.
- The IoController wrappers (with application circuits) are intended to be very magical, and generate all the stuff needed for the microcontroller to work, including programming header, power filtering capacitors, crystals (if frequency-sensitive peripherals like USB and CAN are used), and other passives.
- Rewrite the pin mapper to have the concept of pins as resources, and the ability to have delegated resources (eg, peripherals that are made up of IO pins). This also supports pins with different IO types (eg, some 5v tolerant and some not).
- Support package variations of devices by remapping pin names, both in the PinMappable resource allocator and the VariantPinRemapper for fixed system pins. The base device should define pin and resources by their names (eg, `P0.00`), while the actual devices would provide a mapping to pin numbers.
- In the future, these resource names can be used to autogenerate firmware headers.
- Refactor through all the examples.

Surprisingly, this reduces unit test time by 30%.

Additional infrastructural changes:
- Removes the port dependencies from generators. The old pin mapper was the only user of the API, and the use case is now supported by `port_array.allocated()`
- Resolves #30, removing the last remnants of the old generator API (`Generator.get` removed in favor of defining a generator function with arguments being the solved values)
- `_parent` now refers to the binding parent, and `_lexical_parent` is a separate variable that is used less. Name-of how uses `_parent`, which was needed to support bridged array elements.
- Add `Vector.defined()` to create empty port-arrays.
- Support creating digital ports from both sink and source ports.
- Parts of #97
  - for nRF52840: add requirement USB connected -> VBUS connected
  - add comment about using P1_0 and clearing TRACECONFIG
- Misc cleanup of dead code.

TODOs:
- [x] Deduplicate some more code into IoController.


rohit507 (approved):
LGTM

PR #108, opened 2022-04-04
Continued library cleanup

ducky64:
Hooray, another PR with more lines of code removed than added. Moving more code from frontend to compiler / infrastructure!

Note branch merge target. Depends on #107.

- Add `.connected(... ports ...)` to some blocks, eg `DecouplingCapacitor().connected(gnd, pwr)` as a shorthand to instantiate and connect, with less verbosity than the implicit ports declaration
  - Only supported on some common abstract parts, and only those with typed ports (as opposed to passive ports, since this doesn't compose well with an adapter)
- Change merging blocks (MergedVoltageSource, MergedAnalogSource) to be n-ported using port arrays.
  - Also add a `.connected(...)` method (and change the `merge_from` staticmethod for MergedAnalogSource to be a per-instance `connected` for uniformity.


rohit507 (approved):
Seems reasonable.

PR #109, opened 2022-04-09
Link Array Support

### LINKED ISSUE #20: Block-side Port Arrays

ducky64:
Resolves #20 fully - adds link arrays!
- Infrastructural refactoring to add ConnectionExprUtils, that provides a standardized way to map and manipulate port refs in connected-type constraints
- Infrastructural refactoring to add ConnectedConstraintManager, that centralizes parsing of the different types of connections
- Misc renaming in compiler to be more descriptive and categorical
- Adds LinkArray class in compiler, which includes functionality for expanding from port-array elements and link-elements.
- Complete refactoring of the connect system in the frontend to support array-array connections.
  - Also add support for using the result of connects in further connections - this allows naming connections and tacking more things onto them.
- Unit tests, unit tests, unit tests
- Refactors the RGB LED to use a port-array, which is then connected to the micro with an array connect operation. 1 line instead of 3.
- Refactors some examples to make better use of port-arrays.
- Some example cleanup from past PRs that weren't caught before

TODOs:
- [x] Support array connect allocate <-> link array
- [x] Frontend support, also separate connect construct from connect-flatten construct
- [x] Add flat-connect
- [x] Net renaming in examples

Future cleanups
- [ ] Clean up ExprBuilder/Ref/whatever - maybe standardize on ValueExpr returns?
- [ ] Maybe Ref should also take in `Seq[String]` (used in compiler), in addition to `String*` (used for test construction)
- [ ] Move more logic, especially decoding logic, into ConnectedConstraintManager
- [ ] Change the ElaborateRecord to not carry the constraint names - instead, a separate data structure can map port path -> connected constraints or the ConnectedConstraints data structure
- [ ] Cleanup the array / allocated connection lowering task generation that is just duplicated everywhere - perhaps have a superset function which either skips inapplicable ones, or can control which ones generate?

### DIFF
--- a/compiler/src/main/scala/edg/EdgirUtils.scala
+++ b/compiler/src/main/scala/edg/EdgirUtils.scala
@@ -28,13 +28,15 @@ object EdgirUtils {
         (fn.isDefinedAt(connected.getBlockPort), fn.isDefinedAt(connected.getLinkPort)) match {
           case (true, false) => fn(connected.getBlockPort)
           case (false, true) => fn(connected.getLinkPort)
-          case _ => throw new IllegalArgumentException("block xor link did not match")
+          case (true, true) => throw new IllegalArgumentException("block and link both matched")
+          case (false, false) => throw new IllegalArgumentException("neither block nor link matched")
         }
       case expr.ValueExpr.Expr.Exported(exported) =>
         (fn.isDefinedAt(exported.getExteriorPort), fn.isDefinedAt(exported.getInternalBlockPort)) match {
           case (true, false) => fn(exported.getExteriorPort)
           case (false, true) => fn(exported.getInternalBlockPort)
-          case _ => throw new IllegalArgumentException("exterior xor interior did not match")
+          case (true, true) => throw new IllegalArgumentException("exterior and interior both matched")
+          case (false, false) => throw new IllegalArgumentException("neither interior nor exterior matched")
         }
       case _ => throw new IllegalArgumentException
     }
@@ -45,13 +47,15 @@ object EdgirUtils {
         (fn.isDefinedAt(connected.getBlockPort), fn.isDefinedAt(connected.getLinkPort)) match {
           case (true, false) => connection.update(_.connected.blockPort := fn(connected.getBlockPort))
           case (false, true) => connection.update(_.connected.linkPort := fn(connected.getLinkPort))
-          case _ => throw new IllegalArgumentException("block xor link did not match")
+          case (true, true) => throw new IllegalArgumentException("block and link both matched")
+          case (false, false) => throw new IllegalArgumentException("neither block nor link matched")
         }
       case expr.ValueExpr.Expr.Exported(exported) =>
         (fn.isDefinedAt(exported.getExteriorPort), fn.isDefinedAt(exported.getInternalBlockPort)) match {
           case (true, false) => connection.update(_.exported.exteriorPort := fn(exported.getExteriorPort))
           case (false, true) => connection.update(_.exported.internalBlockPort := fn(exported.getInternalBlockPort))
-          case _ => throw new IllegalArgumentException("exterior xor interior did not match")
+          case (true, true) => throw new IllegalArgumentException("exterior and interior both matched")
+          case (false, false) => throw new IllegalArgumentException("neither interior nor exterior matched")
         }
       case _ => throw new IllegalArgumentException
     }
--- a/compiler/src/main/scala/edg/EdgirUtils.scala
+++ b/compiler/src/main/scala/edg/EdgirUtils.scala
@@ -61,18 +61,20 @@ object EdgirUtils {
     }
 
     // Returns a new connection with the find endpoint replaced with replace. For array connects only.
-    def arrayWithReplacedRef(find: expr.ValueExpr, replace: expr.ValueExpr): expr.ValueExpr = connection.expr match {
+    def arrayUpdateRef(fn: PartialFunction[expr.ValueExpr, expr.ValueExpr]): expr.ValueExpr = connection.expr match {
       case expr.ValueExpr.Expr.ConnectedArray(connected) =>
-        (connected.getBlockPort == find, connected.getLinkPort == find) match {
-          case (true, false) => connection.update(_.connectedArray.blockPort := replace)
-          case (false, true) => connection.update(_.connectedArray.linkPort := replace)
-          case _ => throw new IllegalArgumentException("array block xor link did not match")
+        (fn.isDefinedAt(connected.getBlockPort), fn.isDefinedAt(connected.getLinkPort)) match {
+          case (true, false) => connection.update(_.connectedArray.blockPort := fn(connected.getBlockPort))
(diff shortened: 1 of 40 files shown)


rohit507 (commented):
Seems good. It looks like the examples are getting nicer to read/write too.

rohit507 (approved):

PR #110, opened 2022-04-17
Array-valued expr support through frontend

ducky64:
Adds array-valued expressions to the frontend and literal support to the backend.
- In the frontend, each array type is implemented as its own class (instead of a compositional type). I couldn't figure out how to get mypy to understand compositional types - because it needs both the array element (eg, `FloatExpr`), and the unpacked and castable types (eg, `float` and `Union[float, int, FloatExpr]` respectively). Those three are related (eg, `FloatExpr` is actually `ConstraintExpr[float, Union[float, int, FloatExpr])`) but there doesn't seem to be a way to encode that in the type signature - it can either be `ConstraintExpr[float, Union[float, int, FloatExpr])` or `FloatExpr`, but not both (not cleanly, at least). Intersection types or something like Scala's type members could solve this, but neither seem supported in mypy.
  - Possibly mypy plugins can help, I played around with it but didn't find any useful solutions.
- Refactor ArrayExpr into its own file.
- Lit proto unpacking is now the responsibility of the relevant ConstraintExpr, which makes it more straightforward and consistent to add more ConstraintExpr types (by just extending that class and implementing abstract methods).
- Note that much array support (particularly in the compiler) was built through prior changes, this just adds the final plumbing to make it user-constructable.

Additional changes:
- Made a few more methods in ConstraintExpr into classmethods, and general cleanup.
- Remove dead IR code from frontend. The rest of it is mostly related to TransformUtil, which the netlister still uses (but will be refactored soonTM for multipack devices)


rohit507 (approved):
Seems reasonable.

Issue #87, opened 2021-12-14
Export shouldn't be used with adapters


ducky64:
Export should mainly be used to deduplicate port type data.

Where adapters are needed, this often also needs some kind of forward declaration, so perhaps there should be some other syntax that allows the external port to be defined then connected simply (syntactic sugar) with an adapter.

Issue #81, opened 2021-11-20
Adapters and Bridges


ducky64:
Adapters are currently widely used (common pattern: Passive-typed port for a low-level device like Resistor, with an adapter to a higher-level subcircuit). The current implementation has a few issues:
- Adapters don't work with bridges, since adapters are only defined on the external-facing ports. So VoltageSink can adapt to AnalogSink, but that port is effectively an VoltageSource if viewed from the inside and lacks the `as_analog_sink` method.
- Adapter creation is ad-hoc, currently as functions on the port (like the above `as_analog_sink`). It'd mildly redundant with the adapter block definition.

TODOs
- [ ] from #97, fix the USB resistor parameter propagation in the Raytac part

Main issues / requirements for designing adapters, to explore new options:

## Bridge Support

Obviously, adapters should work nicely with bridges. Currently, the workaround is to manually instantiate the adapter as a block, or to manually instantiate the bridge. Both break abstraction, are verbose, and are bad. Other options:
- Move adapter creation to infrastructure (instead of a per-adapter method defined on each port). A significant overhaul, not sure how this would work. If it's something like `Passive.adapter_to(VoltageSource, voltage_out=2*Volt)` or `Passive.adapter_to(VoltageSource(voltage_out=2*Volt))`, this may limit what type checkers can catch, since the legal set of adapters isn't encoded.
- Define bridged adapter methods on ports, with the same name. For example, define `VoltageSink.as_analog_source(...)`, that would only work (runtime-checked) in the bridge / internal-facing boundary port case.
- Define bridged adapter methods on ports, with a different name. For example, define `VoltageSink.bridged_as_analog_source(...)`. Similar semantics as above, but with another name. The name won't help static typing tools, but might be nice to be more explicit?
- Or, as below, put adapters on the block-side instead of the boundary port side. Unclear if this will work for all cases.

## Adapter Parameterization

Adapters can have different parameter lists: for example, a VoltageSource -> AnalogSource doesn't need additional parameterrsr because it can take the voltage directly from the VoltageSource port, while a Passive -> AnalogSource needs parameters specified in the adapter (because Passive carries no data).

The current structure `VoltageSource.as_analog_source(...)` allows the functions to define an adapter-specific parameter list which can be type-checked. It uses **kwargs to delegate (and deduplicate) the parameter list and parameter handling to the adapter Block's constructor.

## Tooling Support

[...]

PR #111, opened 2022-04-19
Multimeter, revision 2

### LINKED ISSUE #97: Multimeter examples TODOs

ducky64:
Revision of the multimeter example design. Resolves #97 

- Many bugfixes from the first revision (see #97)

- Some major new features for the multimeter example:
  - High resolution delta-sigma ADC
  - Separate analog supply
  - Way cheaper boost converter
  - More common (but still commonly out of stock) MCU
  - Separate SPI for LCD
  - Entire bottom side using JLC components, so many caps I don't need to place...
  - Better layout, I think. 4-sided this time, since it's not as time crunched
  - Test points!
  - Updated fabrication data

- User facing changes
  - Support __getitem__ on port-arrays (from inside only; externally can still only be allocated)

- New parts
  - Elastic connector class, currently 2.54mm and JST PH connector generators
  - Analog switch tree generator (not yet optimized for non-powers-of-2 though)
  - Raytac MDBT50Q module... that is now sold out 😔
  - Tpa2005d1 class-D speaker driver... which is also kinda out of stock
  - ConnectorSpeaker with parameterized specifications, since most speakers are with leads instead of directly board-mount
  - Test points, with value automatically generated from the connected link name
  - Another USB ESD diode, this one also in JLC's parts library

- Infrastructural bugfixes
  - Fix a compiler bug where array expansion dependencies were not properly added
  - Fix a compiler bug where class refinements and subclass default parameter values conflict
  - The inductor selector was ignoring current limits 😱 now the inductors are more reasonably sized
  - Fix a bug with empty array literals
  - Class refinements now apply to all subclasses, post-refinement
  - Support generator access to connected-link parameters
  - nRF52840: only allocate non-low-frequency IOs for digital comms protocols

- Infrastructural upgrades
  - Break each passive (R, C, L) into its own file
  - Refactor parts tables (yet again) - in particular, the sorting logic and part table columns are now defined in the abstract part class instead of being specific (and replicated) by each vendor part table
    - Centralize footprint pinning definitions
    - Clarify vendor name for parts tables, eg DigikeyMlcc, GenericMlcc, JlcCapacitor
  - Add JlcBoardTop with the 3 tooling holes and refinements for JLC parts
  - Much work on JLC part tables
    - Update JLC parts tables to latest, and add a pruned version of the table (with pruning script)
    - Simplify unit tests to be less brittle to table updates
    - Add more JLC parts: FETs, inductors, diodes, zeners
      - Note, the parts tables do not encode inductor frequency, the user must manually verify the inductor frequency and waive the check by refinements.
    - Add a basic parts flag that can be waived (overridden) by refinements
    - Lots of unification and cleanup of code
  - Unify NFet and PFets so N/P are just a parameter in Fet
  - Use ArrayStringExpr (instead of a packed StringExpr) for pin assignments
  - Write allocated pins to a parameter, eventually may be used for firmware header generation
  - In compiler, re-throw exceptions with surrounding context about the running elaboration
  - PeripheralFixedPin now has fixed definitions (it didn't work with the resource allocation system anyways)
  - Remove Digikey table instructions from the main readme, newly downloaded tables have an updated schema that is incompatible anyways

Notes
- KiCad 5.x did not release with the Raytac module footprint, manually [download it from the libraries repo](https://github.com/KiCad/kicad-footprints/blob/master/RF_Module.pretty/Raytac_MDBT50Q.kicad_mod) and install into your RF_Module.pretty library folder. 6.x does include it, but it can't seem to persist netlist references properly #112

TBD list
- ~Maybe fix for #81 and #87~ for a future PR


rohit507 (approved):
seems fine

PR #117, opened 2022-05-24
LED matrix example design

ducky64:
Add charlieplexed LED matrix design, including layout and fabrication data. Good example of generators + array ports.

Library changes
- Add ESP32-C3 (RISCV w/ Wifi/BLE) microcontroller
- Add charlieplexing LED matrix generator with arbitrary row and column configuration
- Add color parameter to LED, as a string (with Color and ColorLike type-alias)
- Add JLC BoM and position file post-processing script, including reformatting and adjusting rotations of some parts

Minor changes
- Add JLC part numbers to more parts used in this design: USB C connector, LD1117
- Add JLC LED generator
- For non-JLC board top, ignore the JLC basic part status
- Add pull-up delay RC, needed for ESP32-C3


rohit507 (approved):
LGTM

PR #122, opened 2022-06-11
Multipack part 2, ToF array example + library updates

ducky64:
Add support (and example - common-anode RGB LED from discrete IndicatorLeds) for multipack parts that share a pin.

Add a ToF (time-of-flight) distance sensor array example board as a second demonstration of array ports (shared I2C + individual shutdown pins - since all the sensors have the same I2C address) and the above multipack. It's got a speaker (so it might work as a rudimentary laser harp or something) and CAN (for kicks and giggles - and I guess to allow chaining several of these).

Compiler changes:
- Fixes an error where multiplying ranges by a negative number doesn't reverse the min/max and instead causes an error
- Add short-circuit evaluation semantics for AND, OR, and IMPLIES - this is needed to handle disconnected ports
  - There needs to be a larger discussion on short-circuit semantics

Infrastructure changes:
- Add `Vector.any_connected()`. It's a bit of a hackfix, there needs to be more thought on dealing with disconnected ports.
- Add NetPackingBlock, that provides a copper connection from the source ports to the destination ports, and asserts (within the netlister) that the sources are connected together
  - This is how multipack parts that share a pin are handled. For example, where the RGB has a single anode, the multipack part takes in the anode of the three LEDs, connects it internally to a PackedVoltageSource, that connects to the internal RGB single anode, then distributes the current draw evenly among the three anode ports.

Library changes:
- Add JLC crystal part
- Add JLC tactile switch
- Add sinking indicator LED (cathode connected to IO, resistor on IO side), needed for RGB packing
- Add RGB packing block from the above
- Add JLC part number to RGB LED
- Add indicator LED array, a simple convenience wrapper to instantiate multiple indicator LEDs
- Move these to electronics_abstract_parts
- Add I2C bridge
  - and a ton of refactoring of DigitalLink to make this work
- Add I2C and CAN test points
- Add passive port to CAN logic-side link - so there is a port to tack the test points onto
- Add SN65HVD230 3.3v-compatible CAN transceiver
- Add VL53L0X distance sensor
- Add JLC part number to STM32F103... surprisingly they still have these in stock, and they're basic parts
- Fix CAN pin def for STM32F103
- Add JLC part number to TPA2005D1 speaker driver
- Add Molex SL PassiveConnector


rohit507 (approved):
LGTM

PR #125, opened 2022-07-22
Design tooltip generator

the6thstar:
Implements description field as part of the BaseBlock class. This field is able to be set in the python code for each part library, and provides a way of accessing a block's variables based on python's f-strings. 

An example use would be -> self.description = self.description = "<b>spec capacitance:</b> {capacitance}F"

Notice that the variable capacitance is wrapped in "{}", which signifies to the IDE that this is a block variable that is to be read.


blinky_skeleton.py
```diff
@@ -1,11 +1,54 @@
...
+    def refinements(self) -> Refinements:
+        return super().refinements() + Refinements(
+        instance_refinements=[
+            (['buck'], Tps561201),
+        ])
 
 
 if __name__ == "__main__":
-  compile_board_inplace(BlinkyExample)
+    compile_board_inplace(BlinkyExample)
```
ducky64, line 54: I generally don't like having test code committed, unless it's part of unit tests. I'd recommend either backing this out or creating a proper automated unit test. It might be a good idea to add a unit test in edg_core that checks the proto output, like some of the other tests there do.

edg_core/Blocks.py
```diff
@@ -168,6 +168,8 @@ def __init__(self) -> None:
 
     self._elaboration_state = BlockElaborationState.init
 
+    self.description = ""
```
ducky64, line 171: A comment here describing what this variable does might be useful.

edg_core/Blocks.py
```diff
@@ -290,6 +292,31 @@ def _populate_def_proto_block_contents(self, pb: BaseBlockEdgirType) -> BaseBloc
 
     return pb
 
+  def _populate_def_proto_description(self, pb: BaseBlockEdgirType) -> BaseBlockEdgirType:
+
```
ducky64, line 296:
```suggestion
```
style nit: don't need newline after function def

edg_core/Blocks.py
```diff
@@ -290,6 +292,31 @@ def _populate_def_proto_block_contents(self, pb: BaseBlockEdgirType) -> BaseBloc
...
+    description = self.description
+    size = len(description)
+
+    stringStart = 0
+
+    for varIndex, char in enumerate(description):
```
ducky64, line 302: I don't think you use `size`? And in which case you only reference `self.description` once and can move it directly into the enumerate?

Also, an alternative architecture would be to use string find, instead of iterating over each character.

edg_core/Blocks.py
```diff
@@ -290,6 +292,31 @@ def _populate_def_proto_block_contents(self, pb: BaseBlockEdgirType) -> BaseBloc
...
+      if char == '{':
+        endIndex = description[varIndex:].find('}')
+        if endIndex != -1:
+
```
ducky64, line 306: style nit: empty line not needed here
```suggestion
```

edg_core/HierarchyBlock.py
```diff
@@ -338,6 +338,8 @@ def _def_to_proto(self) -> edgir.HierarchyBlock:
         assert not port._get_initializers([port_name]), f"connected boundary port {port_name} has unexpected initializer"
     pb = self._populate_def_proto_port_init(pb)
 
+    pb = self._populate_def_proto_description(pb)
```
ducky64, line 341: style nit: should be consistent with the rest of the function - don't need the extra newline

proto/edgir/elem.proto
```diff
@@ -96,6 +96,14 @@ message HierarchyBlock {
   bool is_abstract = 30;  // true if self_class is abstract, and should error if used in a design
 
   edgir.common.Metadata meta = 127;
+
+  message Phrases {
```
ducky64, line 100: Is there a more descriptive name we can give this? For example. StringInterpolatedDescription?

ducky64 (changes requested):
Looks mostly good! A few minor stylistic issues, a piece of dead code, and perhaps add a unit test for completeness.

blinky_skeleton.py
ducky64: Try to write a unit test in the style of https://github.com/BerkeleyHCI/PolymorphicBlocks/blob/master/edg_core/test_block_eltdict.py - create a block with the description field, and check that the proto is as you would expect. You can't do an end-to-end test at the Python level since that requires IDE code, but you can test it was far as Python codes, which is generating protobuf.

edg_core/Blocks.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
```
ducky64, line 3:
```suggestion
```

Dead code? doesn't appear to be referenced elsewhere

edg_core/Blocks.py
```diff
@@ -168,6 +202,8 @@ def __init__(self) -> None:
 
     self._elaboration_state = BlockElaborationState.init
 
+    self.description = DescriptionString("")   # additional string field to be displayed as part of the tooltip for blocks
```
ducky64, line 205: Would it make sense to distinguish between none and empty? For example
```suggestion
    self.description: Optional[DescriptionString] = None   # additional string field to be displayed as part of the tooltip for blocks
```

edg_core/Blocks.py
```diff
@@ -290,6 +326,13 @@ def _populate_def_proto_block_contents(self, pb: BaseBlockEdgirType) -> BaseBloc
 
     return pb
 
+  def _populate_def_proto_description(self, pb: BaseBlockEdgirType) -> BaseBlockEdgirType:
+    description = self.description
+    if isinstance(description, DescriptionString):
```
ducky64, line 331: I'd prefer an assert here, generally I'd prefer to fail noisily on bad input (so the user at least knows something is wrong) than silently discarding data.

edg_core/Link.py
```diff
@@ -16,6 +16,7 @@ class Link(BaseBlock[edgir.Link]):
   def __init__(self) -> None:
     super().__init__()
     self.parent: Optional[Port] = None
+    self.description = DescriptionString("")   # additional string field to be displayed as part of the tooltip for links
```
ducky64, line 19: Is this needed here, since it's already in BaseBlock?

edg_core/test_block_description.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class DescriptionBlockProtoTestCase(unittest.TestCase):
+    def test_description(self):
+        pb = DescriptionBlock()._elaborated_def_to_proto()
+
+        self.assertEqual(len(pb.description), 3)
+
+        self.assertEqual(pb.description[0].text, 'Test variable = ')
+        self.assertEqual(pb.description[1].param.unit, 'Units')
```
ducky64, line 23: Can you also check the reference?

ducky64 (commented):
Mostly looks good, just a few more minor cleanups.

ducky64 (approved):
lgtm!

PR #136, opened 2022-08-15
Improve description strings in electronics libraries

ducky64:
Expands description strings to more classes, also unifies the actual_* parameters (eg, actual_resistance) into abstract base classes so the full description string can be in the base classes.

And some opportunistic code cleanup.

Also some TODOs that could be suggestions for future extensions to the description string system. Perhaps also short description string that ends up in the title, eg ending up with "Capacitor 25uF 100V".


the6thstar (approved):
Looks good!

PR #137, opened 2022-08-16
Bump dependency versions

ducky64:
An outdated version of protoc seems to not be happy with ARM Macs


TayKaiJun (approved):
Works on Macos 12.5

PR #138, opened 2022-08-16
Fix type error, improve booling error message

ducky64:
Otherwise things die on Python 3.8 and below it seems...


TayKaiJun (approved):
Works now!

PR #140, opened 2022-08-17
Refactor PassivePort adapter syntax, add FPC supertype, add OLED

ducky64:
New subcircuits are untested and not yet used, but planned to be in an upcoming board.

Main infrastructure change is the change in PassivePort adapter syntax. Previously, it would have been `port.as_ground()` or `port.as_voltage_sink(... params ...)`, but that's now `port.adapt_to(Ground())` and `port.adapt_to(VoltageSink(... params ...))`, which is more compositional and supports convenience constructors like `DigitalSink.from_supply(...)` without duplicating that in PassivePort.

This adds an abstract superclass for FPC connectors and refactors the existing 8-pin LCD to use this. Also adds a mini OLED display that uses an 15-pin FPC connector. This was the use case for better adapters for Passive - because it uses an internal block with Passive ports instead of a footprint directly it needs to adapt those to the typed (eg, VoltageSink) ports.

Additional minor changes
- Delete unused Port.adapter_types
- Fix description string for DigitalLink
- Add input_threshold_factor option for DigitalSink.from_supply(...)

### DIFF
--- a/edg/BoardTop.py
+++ b/edg/BoardTop.py
@@ -20,6 +20,7 @@ class BaseBoardTop(DesignTop):
 
         (IndicatorSinkLed, IndicatorSinkLedResistor),
 
+        (Fpc050, HiroseFh12sh),
         (UsbEsdDiode, Tpd2e009),
         (TestPoint, Keystone5015),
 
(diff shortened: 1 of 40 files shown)


ducky64:
Merging now since this is gating an example board.

PR #147, opened 2022-09-04
Continued compiler refactoring, towards partial compilation

ducky64:
A variety of refactorings and restructurings that enable partial compilation, particularly adding a hard cut at leaving a block unelaborated and unexpanded.

- Overall, block elaboration is no longer dependent on sub-blocks being expanded or elaborated (as previously)
  - Port checking is done using the sub-block library type, instead of expanding the sub-block directly
- Split expanding blocks and elaborating blocks
  - Expanding is necessary to get the generator dependencies, and elaboration happens once generators (if any) are ready to run
- Refactor ConstProp to resolve CONNECTED_LINK symbolically (the CONNECTED_LINK prefix is rewritten to the link path)
  - Having the connected link is no longer necessary for expanding connections.
  - ConstProp also handles connection resolution to connected link
  - Prior, there was an implicit dependency for connections on link-side ports being expanded (as a result of the connected link being elaborated), this is now an explicitly dependency on the port being elaborated
  - Connected link resolution is now a node in the ConstProp dependency graph
  - Unify all expr evaluation into update(), the initial add in addAssignExpr has no dependencies, just kicks off the update(). Requires the container path and name on all of them (a separate opt-in implicit provides defaults, for writing tests only).
- IS_CONNECTED only available for top-level ports.
  - For exporting to a bundle element, IS_CONNECTED propagates inwards from the bundle's container port
- Restructure how link array connections work (removes undirected equality in favor of directed assignment):
  - Eliminate the undirected ELEMENTS equality in favor of first-assign from block port side ELEMENTS to link ELEMENTS, followed by an assertion that all the block port side ELEMENTS are equal to link ELEMENTS. This simplifies link arrays to use directed assignment only, and avoids brittleness from potential overassignment.
  - Array-connected block side ports ALLOCATED is set from link ELEMENTS. Blocks that are not reactive to the incoming connection size may ignore ALLOCATED. (previously, this was an undirected equality constraint, this is now directed assignment)
- Add partial compilation directive for the Compiler. Block is implemented, param is just a placeholder for now. Future PRs will implement forking a partially-compiled design and some kind of resumption.
- Refactor ConstProp add directed assignment variations (equality, literal value) to delegate to the same underlying method
  - Argument list changes to make them consistent(ish)
- Remove SourceLocator, which never was used anyways, especially since we also are propagating name and container information
- Misc code improvements

HDL library fixes:
- Disallow connected_link references in generator arguments
- Simplifies test point to only take a string name, wrappers can automatically fill that from the link name
- Simplifies I2C bridge to add a pullup flag in the master (defaulting to false), instead of the bridge directly assigning to a connected link.

Potential future refactoring:
- Remove assertions in favor of a transform that checks assertions (redundant state in Compiler)


rohit507 (approved):
There's a bit of a code code smell with how the IS_CONNECTED parameter bubbles down the hierarchy (from the enclosing bundle to it's children) rather than how everything else bubbles information up the hierarchy. 

I suspect it'll lead to a bunch of annoying special casing in the future, especially if someone wants to compose generators. Specifically if they're trying to build a generator for a bundle out of generators for the constituent ports. Thing is, that's a mostly non-sensical use case for EDG as it is and it'll take having multiple domains (e.g. software, FPGA logic, or something analog synthesis-y) in the loop before it becomes an issue. 

As it is, it'll probably not cause trouble in the short term. Hence the approval.

ducky64:
I'm not sure I completely understand the comment.

To add a bit of concreteness, an example would be how the I2C pull up block is structured: the I2C pullup block (which has a I2C typed port) contains two pull up resistors (each of which has a Digital typed port on the IO side). The pullup block exports the Digital typed IO on the resistor, to the relevant boundary I2C port's digital elements (bundle children). With this, connection flows from the hierarchy out-to-in - the inner port being exported inherits the outer port, even if that outer port is a bundle element. Overall, I think the rule for IS_CONNECTED is now "is there a CONNECTED_LINK" (so even if the port was exported, if that export ultimately goes nowhere at the top level, the port would be marked not-connected).

We also don't have generators on ports right now, though generators on blocks can depend on the connected-ness of their boundary ports as generator parameters.

PR #150, opened 2022-09-07
Ki cad greyboxing

nnguyendinh:
Implements the KiCadSchematicBlock, which defines functions to parse a KiCad netlist file and output a connected circuit in edg. As of now, ports must be made by the user and the IDE will match the port names to net labels in the KiCad netlist to connect them. 

To import a KiCad netlist, the user will create a KiCadSchematicBlock like so:
```
class KiCadBlock(KiCadSchematicBlock):
    def __init__(self) -> None:
        super().__init__()

        self.PORT_A = self.Port(Passive())

    def contents(self) -> None:
        super().contents()

        self.import_kicad("INSERT_FILE_PATH_HERE")
```

Further implementation would allow direct parsing of KiCad schematic files instead of a netlist file, and also reduce the responsibility of the user in creating the ports to exactly match the KiCad file.


electronics_model/KiCadSchematicBlock.py
```diff
@@ -0,0 +1,52 @@
...
+        for net_num, net in enumerate(netlist.nets):
+            portlist = []
+            for pin in net.pins:
+                component = getattr(self, pin.ref)
+                if isinstance(component, Resistor):
+                    if pin.num == "1":
+                        portlist.append(component.a)
+                    else:
+                        portlist.append(component.b)
+                if isinstance(component, Capacitor):
```
ducky64, line 25: consider making this an `elif` - it shouldn't be possible for both cases to be true

electronics_model/test_kicad_import.py
```diff
@@ -0,0 +1,90 @@
...
+        super().contents()
+
+        self.import_kicad(os.path.join(os.path.dirname(__file__), "resources", "test_kicad_import.net"))
+
+class BlinkyExample(BoardTop):
+    def __init__(self):
+        super().__init__()
+
+        self.kicad = self.Block(KiCadBlock())
+        print("path = " + os.path.dirname(__file__))
```
ducky64, line 26: Generally discouraged to have prints in unit tests - unit tests should be fully automated

electronics_model/test_kicad_import.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.PORT_A = self.Port(Passive())
+
+    def contents(self) -> None:
+        super().contents()
+
+        self.import_kicad(os.path.join(os.path.dirname(__file__), "resources", "test_kicad_import.net"))
+
```
ducky64, line 20: two newlines between classes I think

electronics_model/test_kicad_import.py
```diff
@@ -0,0 +1,90 @@
...
+class BlinkyExample(BoardTop):
```
ducky64, line 21: Is this used? Can this be deleted?

electronics_model/test_kicad_import.py
```diff
@@ -0,0 +1,90 @@
...
+
+        expected_conn = edgir.ValueExpr()
+        expected_conn.connected.link_port.ref.steps.add().name = 'PORT_A_link'
+        expected_conn.connected.link_port.ref.steps.add().name = 'passives'
+        expected_conn.connected.link_port.ref.steps.add().allocate = ''
+        expected_conn.connected.block_port.ref.steps.add().name = 'R1'
+        expected_conn.connected.block_port.ref.steps.add().name = 'a'
+        self.assertIn(expected_conn, self.pb.constraints.values())
+
+        expected_conn = edgir.ValueExpr()
```
ducky64, line 84: For simplicity and to keep tests orthogonal and independent it may be better to have a circuit where a bridge doesn't need to be inferred - practically this means that only one pin is connected to a boundary port (PORT_A here)

ducky64 (changes requested):
This look good as a proof of concept, and I think this is an appropriate scope to merge especially in the spirit of frequent and fast PRs!

I added a few comments, some of them are style nits and some of them are functional improvements. While PRs don't need to be a fully complete feature, I think the stuff implemented should be clean.

Additional changes requested:
- Can you add the source schematic?
- Can you also add `kinparse` to the Pipfile? (this allows automatic fetching of dependencies in some workflows)

For the future (a different PR):
- Implement parsing of SI prefixes, eg MOhm - there may be some shared code with the parts table parser
- Support a more concise version of defining supporting components - preferably in a defined data structure (such as `NamedTuple`) that combines the part parsing and pin mapping together and is user extensible (so like a dict that can be appended to without modifying infrastructure code, instead of if statements that can only be modified by changing this code)
- Support instantiating components with Python code in the value field, and defining the mapping in the instantiated block itself, preferable with shared code with the case where it recognizes the component type from the KiCad desc

blinky_skeleton.py
```diff
@@ -1,10 +1,22 @@
 from edg import *
+from electronics_model.KiCadSchematicBlock import KiCadSchematicBlock
 
 
+class KiCadBlock(KiCadSchematicBlock):
```
ducky64, line 5: Can you back out the changes to blinky_skeleton? This is meant as an example starter and should be clean.

It's fine to use it for quick and dirty testing (though I think using the unit test system is better), but quick and dirty code should not be merged back in as final code.

ducky64 (changes requested):
Almost there - just back out the blinky_skeleton testing code and it's good!

blinky_skeleton.py
nnguyendinh: Oops sorry forgot that I left that in. Just changed it back to what's in the master branch.

ducky64 (approved):
lgtm!

PR #163, opened 2022-10-09
Rewrite getting started tutorial

ducky64:
Changes:
- Update the tutorial for new features: port arrays, abstract microcontroller.
- Integrate the IDE tutorial into the main getting started tutorial. The tutorial in the IDE repository will be deleted and redirect here instead.
- Separate the library building section into its own document.
- Add an advanced library building section covering defining blocks with port arrays.
- Add an advanced top-level section covering multipack parts (but not their definition).


getting-started.md
```diff
@@ -35,406 +38,669 @@ While this getting started guide is meant to be self-contained, you may also fin
...
-self.mcu = self.Block(Nucleo_F303k8())
+self.usb = self.Block(UsbCReceptacle())
+self.mcu = self.Block(Stm32f103_48())
 self.led = self.Block(IndicatorLed())
 ```
 > `self.Block(...)` creates a sub-block in `self` (the current hierarchy block being defined).
 > It must be assigned to an instance variable (in this case, `mcu`), which is used as the name sub-block.
 
+> You can also insert blocks through graphical operations in the IDE.
+> 1. Start by selecting the insert location in code, by setting the caret to the end of `super().contents()`.
```
PierceCCH, line 168: Maybe we can emphasize and explain that the caret must be at that specific position.

getting_started_advanced.md
```diff
@@ -0,0 +1,276 @@
...
+  ...
+  def generate(self, count: int) -> None:
+    for i in range(count):
+      self.ios.append_elt(DigitalSink.empty())
+      ...
+```
+
+> Port array's `.append_elt(...)` takes in the same arguments as `self.Port(...)`.
+> 
+> To mark a port array as explicitly having no elements, use `.defined()`.
```
PierceCCH, line 137: maybe there can be an example of how this is used

TayKaiJun (approved):
Minor comments for Tutorial documentations

getting_started_advanced.md
```diff
@@ -0,0 +1,276 @@
...
+    ...
+    with self.implicit_connect(
+            ...
+    ) as imp:
+      ...
+      self.led = imp.Block(LedArray(4))
+      self.connect(self.mcu.gpio.request_vector('led'), self.led.ios)
+```
+
+Or, since the new block has both implicit scope and chain tags:
```
PierceCCH, line 181: I didn't really understand this line. Maybe an additional explanation of implicit scope and chain tags can be added.

PierceCCH (approved):

getting_started_advanced.md
ducky64: good idea, added

getting_started_advanced.md
ducky64: Clarified by referencing the `Input` tag, I'm hoping that will jog people's memory enough.

In the longer term we should probably write a reference document for these concepts, and link them from here.

PR #164, opened 2022-10-13
Robot Driver Cleanup

ducky64:
Makes robotdriver2 the only robotdriver, since the original had issues (mostly around connecting peripheral IOs through an IO expander, which doesn't work well with Arduino libraries that expect to be able to digitalWrite things). Also updates the code to rename allocate to request.


TayKaiJun (approved):
Looks good to go!

PR #165, opened 2022-10-14
Address common errors from the tutorial

ducky64:
Add error message examples and resolutions for Python-dependencies-not-found and submodule-not-cloned


PierceCCH (approved):
looks good to me

PR #167, opened 2022-10-23
Mechanical keyboard components

ducky64:
- Add SK6812Mini-E, reverse-mount Neopixel LED which is commonly used for per-key lighting. Add untested footprint.
- Add Kailh switch socket, with an external footprint library dependency.
- Refactor the WS2812 to extend a base Neopixel class, so the NeopixelArray can be used with the SK6812.
- Add blank keyboard top-level board project.
- Tested that the components do netlist correctly.


TayKaiJun (commented):
Ran into a compiler error for RobotDriver board. I don't think it was due to the change in the abstract Neopixel class though. Here's the error log:
image

Otherwise everything else lgtm!

ducky64:
Interesting, it thinks something is weird with... the inductor? 😮 

I think this might be the IDE's HDL compiler cache being in a bad state, I've seen that happen with changing branches though I'm not exactly sure what's going on here. If you empty the cache and rebuild do you still get the same error?

TayKaiJun:
> 

Okay yes it was the IDE cache messing up again.

TayKaiJun (approved):
good to go!

PR #174, opened 2022-11-10
Improve setup documentation and debuggability

ducky64:
- catches and stores errors if the Python subprocess fails to start, so errors can be reported through the normal flow instead of just crashing immediately - in particular for unit tests, this means the error message is reported, instead of directly printed and immediately clobbered by the test failure spam that doesn't include useful diagnostics
- add Deprecated pip package to README
- update the submodule not cloned error to use the new sub-project name
- add python-is-python3 issue and solution for Linux
- suggest debugging with just one test instead of all of them
- suggest JRE/JDK versions
- runs natively on major OSes

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Runs natively on Windows, Linux, and Mac.
 
 The packages can be installed using `pip`:
 ```python
-pip install protobuf kinparse Deperecated
+pip install protobuf kinparse Deprecated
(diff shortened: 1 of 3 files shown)


TayKaiJun (approved):
good to go

PierceCCH (approved):
lgtm

PR #175, opened 2022-11-21
Order map elements in proto

### LINKED ISSUE #113: Eliminate unordered map in proto
It's a lot of hacks to append ordering data (and often imperfectly so), it should just be a list of key-value pairs

ducky64:
Changes the proto from `map<string, V>` to use `repeated` of `(name, V)` sub-messages. Resolves #113 

Implementation notes:
- In Scala, this defines an implicit .toPb on SeqMap[Str, ElemType], and an implicit .toSeqMap (when a SeqMap specifically is needed) and .asPairs (to compose with Seq methods) on Seq[ProtoType] (for the name/value pair sub-messages).
- Proto construction utils use ordered SeqMap instead of unordered Map,
- In Python, add a edgir.add_pair (to add a NamedPair with specified name to the repeated field, returning the value to be populated) util, since Python unfortunately does not support nice-looking implicits.
- Also add a edgir.pair_get_opt (for a list of NamedPair, return the one of specified type or None) and .pair_get (requires the index exists)
- Lots of updating test code.
- Eliminate naming order metadata (obsoleted).
- Refactor frontend to not generate metadata field at all if there is no metadata.
- Frontend internals: change ordering in which constraints are generated, required connects then port inits then param inits.
- Refactor FootprintBlock to use fp_pinning param instead of metadata, as part of a general effort to clean up metadata.
- Remove footprint proto test, in favor of just keeping the footprint to netlister test instead of testing internal details. Also the array literal test isn't straightforward =(
- Delete some no-longer-used metadata.

Future refactorings:
- Refdes annotation transform, separate from netlister
- Ensure netlister is deterministic
- Eliminate usage of metadata for netlist-related data, instead using param and the type hierarchy


rohit507 (approved):
lgtm

PR #178, opened 2022-12-03
Separate refdesing from netlisting, refinement-pass infrastructure

ducky64:
... allowing the refdes to be visible as a parameter, eg for the IDE tree view and future PDFing.

This introduces a new HDL-interface construct (with corresponding proto), a refinement pass that takes a design and a list of existing solved values, and returns new solved values (here, the refdes)

Other minor refactorings:
- In unit tests, prefer NetlistBackend (instead of NetlistTransform), at least where we aren't examining the intermediate output
- DRY out the netlisting code in some unit tests in NetlistTestCase.generate_net


rohit507 (approved):
LGTM

PR #181, opened 2022-12-13
gitignore generated PDF files

ducky64:
Since we now have generated .pdf files, they probably shouldn't be committed just like the generated .net (and other) files.


TayKaiJun (approved):
This is great

PR #194, opened 2023-01-09
Bom

ParthPandhare:
Most of the Bom code is in 'BomBackend.py' , with the generation taking place in 'test_kicad.py'.

Considerations for the future:
- naming BoM file based on the schematic name
- combining BoM generation with netlist generation


ducky64 (changes requested):
Cool, thanks!

It looks like a different branch (the one with FPGAs) got merged in here, let's look into separating out the changes so this PR is self-contained. Also, changes to blinky_skeleton shouldn't be pushed, that's purely a getting started skeleton. Try to keep the PR diff contained to only the feature under development.

Otherwise, looks like it works and the core logic is pretty solid (so it's functionally good)! I have some stylistic suggestions in the inline comments that could make the code cleaner, see if you like those. Also, file IO should be avoided in the BackBackend interface, though granted that wasn't documented anywhere.

electronics_model/BomBackend.py
```diff
@@ -0,0 +1,67 @@
...
+        self.part_number = part_number
+        self.datasheet = datasheet
+
+
+class GenerateBom(BaseBackend):      # creates and populates .csv file
+    def run(self, design: CompiledDesign) -> List[Tuple[edgir.LocalPath, str]]:
+        bom_list = BomTransform(design).run()
+        name = os.path.splitext(os.path.basename(__file__))[0] + '_bom.csv'
+        with open(name, 'w', newline='') as f:
+            fieldnames = ['Ref Des', 'Quantity', 'Value', 'Footprint', 'Manufacturer', 'Part Number', 'Datasheet']
```
ducky64, line 26: Also, consider making the column headers / overall file format consistent with one of the KiCad BoM export formats, so any tools that use that format can use our output. I'd prefer using the KiCad PCB BoM generator as a simple reference, since there's so many options for the schematic BoM generator and this is what we've been using.

ducky64 (changes requested):
Looking better, though some more suggestions inline.

Still on the todo list:
- git magic to split out this feature
- avoid committing generated files - maybe we should expand the .gitignore rules

electronics_model/BomBackend.py
```diff
@@ -0,0 +1,48 @@
...
+
+import edgir
+from edg_core import BaseBackend, CompiledDesign, TransformUtil
+
+import csv
+
+
+class BomItem(NamedTuple):
+    footprint: str
+    value: str
```
ducky64, line 12: Can you also add a field for JLC part number?

The column name in the CSV should be `JLCPCB Part #`, and the part parameter would be `lcsc_part`. Like value, this should be empty if None.

ducky64 (changes requested):
This looks mostly good! I added a few most suggestions in the inline comments along with a feature request (JLC part number). That feature can be part of a separate PR if you'd like, we can merge this one without it.

In addition, please remove these generated files from this PR (in general it's not good practice to commit generated files):
- BlinkyExample.pdf
- test_kicad_*.csv
- revert the changes to blinky_skeleton.py, it should be an empty skeleton design

electronics_model/BomBackend.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class GenerateBom(BaseBackend):      # creates and populates .csv file
+    def run(self, design: CompiledDesign, args=None) -> List[Tuple[edgir.LocalPath, str]]:
+        if args is None:
+            args = {}
+        bom_list = BomTransform(design).run()
+
+        bom_string = io.StringIO()
+        csv_data = ['Id', 'Designator', 'Package', 'Quantity', 'Value', 'Designation']  # populates headers
```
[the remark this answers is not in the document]
ParthPandhare: oh my bad, I completely forgot to switch it when (I was testing with Value temporarily). It's been fixed!

electronics_model/test_kicad.py
```diff
@@ -5,22 +5,33 @@
...
     compiled = ScalaCompiler.compile(design)
     compiled.append_values(RefdesRefinementPass().run(compiled))
-    return NetlistBackend().run(compiled)[0][1]
+
+    return NetlistBackend().run(compiled)[0][1], GenerateBom().run(compiled)[0][1]
 
   def test_basic_kicad(self):
     with open(os.path.splitext(os.path.basename(__file__))[0] + '_basic.net', 'w') as f:
-      f.write(self.generate_net(test_netlist.TestBasicCircuit))
+      f.write(self.generate_net(test_netlist.TestBasicCircuit)[0])
```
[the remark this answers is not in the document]
ParthPandhare: updated this

electronics_model/BomBackend.py
ParthPandhare: done

ducky64 (approved):
lgtm!

PR #201, opened 2023-01-19
CI: test for Python 3.9 and 3.11

ducky64:
Adds Python 3.9 and 3.11 tests to automated PR tests.

Fixes a typing bug for 3.11, where our assert_type conflicts with something - presumably a base Python construct added in 3.11. It's now named assert_cast.

Pre-Python 3.9 isn't supported since 3.9 introduces str.removeprefix and str.removesuffix, which we use

Other changes:
- Rename Exceptions.py to HdlUserExceptions.py, which is more descriptive and less likely to collide
- Style: add trailing newline in a documentation file. Partially to trigger a re-run of tests.


ParthPandhare (commented):
Generating a schematic works (with blinky skeleton) however I'm still having trouble with the tests. When I tried to run NetlistTestCase (test_kicad.py) it still showed me 3 errors but when I used the terminal to run the unittest it worked fine.

PR #211, opened 2023-02-23
Price tool

ParthPandhare:
Can you look over the objective function in python?


electronics_model/PriceGetter.py
```diff
@@ -0,0 +1,69 @@
...
+import re
+
+
+class GetPrice:
+    def run(lcsc_part_number: str, quantity: int) -> float:
+        cur_path = os.path.dirname(__file__)
+        parts_library = os.path.relpath(
+            '..\\electronics_lib\\resources\\Pruned_JLCPCB SMT Parts Library(20220419).csv', cur_path)
+
+        price_string: str = 'NA'
```
ducky64, line 16:
```suggestion
        price_string: Optional[str] = None
```
While this will work, consider using the Optional type instead so it's strongly typed (the missing-value type is different) and the type system can help make sure all cases are considered.

electronics_model/PriceGetter.py
```diff
@@ -0,0 +1,69 @@
...
+            csv_reader = csv.DictReader(csv_file, 'r')
+            for row in csv_reader:
+                if row['LCSC Part'] == lcsc_part_number:
+                    price_string = row['Price']
+                    break
+            if price_string == 'NA':
+                return -1
+            csv_file.close()
+
+        temp = price_string.split(',')  # temp = list of all the qty ranges & prices per range
```
ducky64, line 27: Style: consider giving this a more descriptive name, maybe prices_by_quantity or something?

electronics_model/PriceGetter.py
```diff
@@ -0,0 +1,69 @@
+from typing import List, Tuple, Dict
+
+import edgir
+from edg_core import BaseBackend, CompiledDesign, TransformUtil
+import os
+import csv
+import re
+
+
+class GetPrice:
```
ducky64, line 10: Does this need a class? Can this just be a function in `GeneratePrice` - I don't think this needs any dedicated state?

ducky64 (commented):
This look good! I added some suggestions in the review comments, these are mostly style and robustness (resilience to malformed / unexpected input and failing noisily so we can debug it if things aren't as we expect) improvements. Let me know if you have any questions.

We should discuss integration into the IDE next.

electronics_model/PriceGetter.py
ParthPandhare: got it, fixed

electronics_model/PriceGetter.py
ParthPandhare: done

electronics_model/PriceGetter.py
```diff
@@ -0,0 +1,69 @@
...
+        price_string: str = 'NA'
+        with open(parts_library, 'r', newline='') as csv_file:
+            csv_reader = csv.DictReader(csv_file, 'r')
+            for row in csv_reader:
+                if row['LCSC Part'] == lcsc_part_number:
+                    price_string = row['Price']
+                    break
+            if price_string == 'NA':
+                return -1
+            csv_file.close()
```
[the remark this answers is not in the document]
ParthPandhare: you're right, it's redundant. I got rid of it

electronics_model/PriceGetter.py
```diff
@@ -0,0 +1,69 @@
...
+import os
+import csv
+import re
+
+
+class GetPrice:
+    def run(lcsc_part_number: str, quantity: int) -> float:
+        cur_path = os.path.dirname(__file__)
+        parts_library = os.path.relpath(
+            '..\\electronics_lib\\resources\\Pruned_JLCPCB SMT Parts Library(20220419).csv', cur_path)
```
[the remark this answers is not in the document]
ParthPandhare: makes sense, I changed the file location to electronics lib

electronics_model/PriceGetter.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def run(self) -> Dict[str, int]:
+        self.transform_design(self.design.design)
+        return self.part_list
+
+
+class GeneratePrice(BaseBackend):      # creates and populates .csv file
+    def run(self, design: CompiledDesign, args=None) -> List[Tuple[edgir.LocalPath, str]]:
+        if args is None:
+            args = {}
```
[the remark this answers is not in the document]
ParthPandhare: added the assert

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+        self.transform_design(self.design.design)
+        return self.part_list
+
+
+class GeneratePrice(BaseBackend):
+    # part_number -> [ [lower-bound_1, price_1], [lower-bound_2 (and the previous upperbound), price_2], ... ]
+    PRICE_TABLE = Dict[str, List[Tuple[int, float]]] = None
+
+    def generate_price_table(self):
+        if GeneratePrice.PRICE_TABLE is not None:
```
ducky64, line 31:
```suggestion
        if GeneratePrice.PRICE_TABLE is None:
```
Wait, it should only create the table if it is `None` right?

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+    def run(self) -> Dict[str, int]:
+        self.transform_design(self.design.design)
+        return self.part_list
+
+
+class GeneratePrice(BaseBackend):
+    # part_number -> [ [lower-bound_1, price_1], [lower-bound_2 (and the previous upperbound), price_2], ... ]
+    PRICE_TABLE = Dict[str, List[Tuple[int, float]]] = None
+
+    def generate_price_table(self):
```
ducky64, line 30:
```suggestion
    @classmethod
    def get_price_table(cls) -> Dict[...]:
```

Consider implementing this as `get_part_table`, such that the user calls this to get the table instead of directly accessing the variable, and if the table hasn't already been parsed, does the parsing. This way, the function user doesn't need to remember to call a separate generate method, the function is correct without additional preconditions.

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+        if GeneratePrice.PRICE_TABLE is not None:
+            cur_path = os.path.dirname(__file__)
+            parts_library = os.path.relpath('resources/Pruned_JLCPCB SMT Parts Library(20220419).csv', cur_path)
+            with open(parts_library, 'r', newline='') as csv_file:
+                csv_reader = csv.DictReader(csv_file, 'r')
+                for row in csv_reader:
+                    full_price_list = row['Price']
+                    price_and_quantity_groups = full_price_list.split(',')
+                    value: List[Tuple[int, float]]
+                    for i in range(len(price_and_quantity_groups)):
```
ducky64, line 40: consider directly iterating over the list elements!

```suggestion
                    for price_and_quantity in price_and_quantity_groups:
```

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+    def generate_price_table(self):
+        if GeneratePrice.PRICE_TABLE is not None:
+            cur_path = os.path.dirname(__file__)
+            parts_library = os.path.relpath('resources/Pruned_JLCPCB SMT Parts Library(20220419).csv', cur_path)
+            with open(parts_library, 'r', newline='') as csv_file:
+                csv_reader = csv.DictReader(csv_file, 'r')
+                for row in csv_reader:
+                    full_price_list = row['Price']
+                    price_and_quantity_groups = full_price_list.split(',')
+                    value: List[Tuple[int, float]]
```
ducky64, line 39:
```suggestion
                    value: List[Tuple[int, float]] = []
```
type annotations are purely annotations (nonfunctional), it still needs a value

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+                    for i in range(len(price_and_quantity_groups)):
+                        temp = price_and_quantity_groups[i].split(':')
```
ducky64, line 41: could use a more descriptive variable name?

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+                        quantity_range = temp[0].split('-')
+                        value.append((int(quantity_range[0]), float(temp[1])))
```
ducky64, line 43: might be a good idea to assert on the lengths of the results of the `split`s, just in case there's weird formatting we missed - better fail noisily than give a bad result silently

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+
+                    # sorts each value for PRICE_TABLE by lower-bounds, which is the first int in the Tuple:
+                    sorted(value)
+                    GeneratePrice.PRICE_TABLE[row['LCSC Part']] = value
+
+    def generate_price(self, lcsc_part_number: str, quantity: int) -> float:
+        full_price_list = GeneratePrice.PRICE_TABLE[lcsc_part_number]
+        for i in range(len(full_price_list)):
```
ducky64, line 51: also consider iterators here - you actually might be able to unpack the tuple in the iterator:
```suggestion
        for (min_quantity, quantity_price) in full_price_list:
```

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+            if i == len(full_price_list) - 1:   # if you reached the last group
+                return quantity * (full_price_list[i])[1]
+            if quantity < (full_price_list[i + 1])[0]:
+                return (quantity * full_price_list[i])[1]
+
+    def run(self, design: CompiledDesign, args=None) -> List[Tuple[edgir.LocalPath, str]]:
+        assert(args is None)
+        args = {}
```
ducky64, line 59:
```suggestion
```

Unnecessary?

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+    def generate_price(self, lcsc_part_number: str, quantity: int) -> float:
+        full_price_list = GeneratePrice.PRICE_TABLE[lcsc_part_number]
+        for i in range(len(full_price_list)):
+            if i == len(full_price_list) - 1:   # if you reached the last group
+                return quantity * (full_price_list[i])[1]
+            if quantity < (full_price_list[i + 1])[0]:
+                return (quantity * full_price_list[i])[1]
+
+    def run(self, design: CompiledDesign, args=None) -> List[Tuple[edgir.LocalPath, str]]:
+        assert(args is None)
```
ducky64, line 58:
```suggestion
        assert args is None
```
`assert`s are weird in Python, they're actually not a function and I think the parens notation is discouraged.

Alternatively, you can also consider `assert not args` - an empty collection evaluates to false, so this enables both the None case and the empty case.

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,68 @@
...
+        args = {}
+        price_list = PriceTransform(design).run()
+        total_price = 0
+        self.generate_price_table()
+        for lcsc_part_number, quantity in price_list:
+            price_list += self.generate_price(lcsc_part_number, quantity)
```
ducky64, line 64: 🤔

ducky64 (commented):
Looks better, some more stylistic and robustness suggestions

electronics_lib/PriceGetter.py
ParthPandhare: done!

electronics_lib/PriceGetter.py
ParthPandhare: if I do that, can I still access the next element in full_price_list to get the upper-bound?

electronics_lib/PriceGetter.py
ParthPandhare: not really lol, I removed it

electronics_lib/PriceGetter.py
ParthPandhare: typo, fixed

electronics_lib/PriceGetter.py
ParthPandhare: makes sense, I changed it

electronics_lib/PriceGetter.py
ducky64: Good point - not easily (though some languages have a builtin sliding window iterator which is pretty cool)

But, it might be worth thinking of structuring the data such that you can get everything you need with elementwise iteration, here that might be having the upper bound with the price.

electronics_lib/PriceGetter.py
```diff
@@ -27,24 +27,28 @@ class GeneratePrice(BaseBackend):
...
+                    for price_and_quantity in price_and_quantity_groups:
+                        price_and_quantity_list = price_and_quantity.split(':')
+                        assert len(price_and_quantity_list) == 2
+                        quantity_range = price_and_quantity_list[0].split('-')
+                        assert len(quantity_range) == 1 or len(quantity_range) == 2
+                        value.append((int(quantity_range[0]), float(price_and_quantity_list[1])))
 
                     # sorts each value for PRICE_TABLE by lower-bounds, which is the first int in the Tuple:
                     sorted(value)
                     GeneratePrice.PRICE_TABLE[row['LCSC Part']] = value
```
ParthPandhare, line 50: I'm getting an error here & idk what it is (i tried searching for it but couldn't really understand what's going on): "Cannot find reference '[' in 'None' "

ducky64: I think you need to initialize `GeneratePrice.PRICE_TABLE` to an empty dict, otherwise it's trying to index into the default None. Though that error message is really pretty unhelpful...

Also consider seeing what mypy is telling you, it should be particularly helpful for potential type-based errors like this.

electronics_lib/PriceGetter.py
```diff
@@ -25,47 +26,58 @@ def run(self) -> Dict[str, int]:
...
-    PRICE_TABLE = Dict[str, List[Tuple[int, float]]] = {}
+    PRICE_TABLE: Dict[str, List[Tuple[int, float]]] = {}
 
     @classmethod
     def get_price_table(cls) -> Dict[str, List[Tuple[int, float]]]:
         if not GeneratePrice.PRICE_TABLE:
-            cur_path = os.path.dirname(__file__)
-            parts_library = os.path.relpath('resources/Pruned_JLCPCB SMT Parts Library(20220419).csv', cur_path)
+            current_path = os.getcwd()
+            parts_library = current_path + "\PolymorphicBlocks\electronics_lib\\resources\Pruned_JLCPCB SMT Parts Library(20220419).csv"
```
ducky64, line 35: I'd recommend using the `PartsTable.with_source_dir` utility function like in [JlcPart](https://github.com/BerkeleyHCI/PolymorphicBlocks/blob/master/electronics_lib/JlcPart.py#L48) so the path is relative to this file. `os.getcwd` will probably differ based on whether it's within the IDE or run from the command line as eg blinky_skeleton.py or unit tests.

electronics_lib/PriceGetter.py
```diff
@@ -25,47 +26,58 @@ def run(self) -> Dict[str, int]:
...
             with open(parts_library, 'r', newline='') as csv_file:
-                csv_reader = csv.DictReader(csv_file, 'r')
+                csv_reader = csv.reader(csv_file, dialect='excel')
```
ducky64, line 37: Not sure about `excel` dialect, but typically `encoding='gb2312'` works for these parts tables

electronics_lib/PriceGetter.py
```diff
@@ -25,47 +26,58 @@ def run(self) -> Dict[str, int]:
...
-                    value: List[Tuple[int, float]] = []
-                    for price_and_quantity in price_and_quantity_groups:
-                        price_and_quantity_list = price_and_quantity.split(':')
-                        assert len(price_and_quantity_list) == 2
-                        quantity_range = price_and_quantity_list[0].split('-')
-                        assert len(quantity_range) == 1 or len(quantity_range) == 2
-                        value.append((int(quantity_range[0]), float(price_and_quantity_list[1])))
+                    full_price_list = row[10]
+                    if full_price_list == " ":
+                        print("Part " + row[0] + " is missing from the price list.")
```
ducky64, line 42: Hm, how about just skipping the row (not adding a value to the table)? So that if the part isn't used, we don't see anything, and if the part is used, we can see the errors (missing from price list) with the other errors?

A robust way to parse this might be `if full_price_list.strip():` - it's a non-whitespace-sensitive test for emptyness

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,83 @@
...
+                        for price_and_quantity in price_and_quantity_groups:
+                            price_and_quantity_list = price_and_quantity.split(':')
+                            assert len(price_and_quantity_list) == 2
+                            quantity_range = price_and_quantity_list[0].split('-')
+                            assert len(quantity_range) == 1 or len(quantity_range) == 2
+                            value.append((int(quantity_range[0]), float(price_and_quantity_list[1])))
+
+                    # sorts each value for PRICE_TABLE by lower-bounds, which is the first int in the Tuple:
+                    sorted(value)
+                    GeneratePrice.PRICE_TABLE[row[0]] = value
```
ducky64, line 56:
```suggestion
                    GeneratePrice.PRICE_TABLE[row[0]] = sorted(value)
```
I think `sorted` returns the sorted list, there's a different function to do an in-place sort.

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,83 @@
...
+    PRICE_TABLE: Dict[str, List[Tuple[int, float]]] = {}
+
+    @classmethod
+    def get_price_table(cls) -> Dict[str, List[Tuple[int, float]]]:
+        if not GeneratePrice.PRICE_TABLE:
+            current_path = os.getcwd()
+            parts_library = current_path + "\PolymorphicBlocks\electronics_lib\\resources\Pruned_JLCPCB SMT Parts Library(20220419).csv"
+            with open(parts_library, 'r', newline='') as csv_file:
+                csv_reader = csv.reader(csv_file, dialect='excel')
+                next(csv_reader)
```
ducky64, line 38: Add a comment here? Presumably this is to eat the header row since you're not using DictReader?

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,83 @@
...
+                        print("Part " + row[0] + " is missing from the price list.")
+                        value = ([(0, 0.0), (20000, 0.0)])
+                    else:
+                        price_and_quantity_groups = full_price_list.split(',')
+                        value: List[Tuple[int, float]] = []
+                        for price_and_quantity in price_and_quantity_groups:
+                            price_and_quantity_list = price_and_quantity.split(':')
+                            assert len(price_and_quantity_list) == 2
+                            quantity_range = price_and_quantity_list[0].split('-')
+                            assert len(quantity_range) == 1 or len(quantity_range) == 2
```
ducky64, line 51: A comment here may also be helpful, eg when it's 1 element it means a minimum quantity for some price break, instead of a quantity range

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,83 @@
...
+                            value.append((int(quantity_range[0]), float(price_and_quantity_list[1])))
+
+                    # sorts each value for PRICE_TABLE by lower-bounds, which is the first int in the Tuple:
+                    sorted(value)
+                    GeneratePrice.PRICE_TABLE[row[0]] = value
+        return GeneratePrice.PRICE_TABLE
+
+    def generate_price(self, lcsc_part_number: str, quantity: int) -> float:
+        if lcsc_part_number in GeneratePrice.PRICE_TABLE:
```
ducky64, line 60:
```suggestion
        if lcsc_part_number in self.get_price_table():
```

The general way I'd use this is instead of referring to `GeneratePrice.PRICE_TABLE` directly, always use the wrapper `get_price_table` function. Sure, there will be a mild overhead in calling the function instead of just getting the variable, but it also makes sure that all accesses are correct - there are no preconditions to worry about.

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,83 @@
...
+                    return (quantity * full_price_list[i])[1]
+        else:
+            print(lcsc_part_number + " is missing from the price list.")
+            return 0
+
+    def run(self, design: CompiledDesign, args=None) -> List[Tuple[edgir.LocalPath, str]]:
+        assert not args
+        price_list = PriceTransform(design).run()
+        total_price = 0
+        self.get_price_table()
```
ducky64, line 75: ... and if you use `self.get_price_table()` where the price table is needed, you don't need this line here

electronics_lib/PriceGetter.py
```diff
@@ -0,0 +1,83 @@
...
+
+                    # sorts each value for PRICE_TABLE by lower-bounds, which is the first int in the Tuple:
+                    sorted(value)
+                    GeneratePrice.PRICE_TABLE[row[0]] = value
+        return GeneratePrice.PRICE_TABLE
+
+    def generate_price(self, lcsc_part_number: str, quantity: int) -> float:
+        if lcsc_part_number in GeneratePrice.PRICE_TABLE:
+            full_price_list = GeneratePrice.PRICE_TABLE[lcsc_part_number]
+            for i in range(len(full_price_list)):
```
ducky64, line 62: Generally, I still prefer iterating over the elements so there isn't indexing to worry about, eg `for (quantity, price) in full_price_list:`. One way to make this work might be to initialize the price to some default (perhaps the lowest bracket), then iterate over the list, and since we're assuming it's in order, for every price bracket that the quantity exceeds the minimums for, overwrite that price. Return happens at the end of iteration.

ducky64 (changes requested):
Looking better! A few more suggestions inline.

ducky64 (approved):
lgtm!

PR #247, opened 2023-04-19
Python packaging defs

ducky64:
- Add pyproject.toml
- Refactor paths for packaging:
  - precompiled compiler now in edg_core/resources
  - HDL interface server in its own package, to be runnable with `python -m`
- Update documentation
- Package uploaded to pip


PierceCCH (approved):
completed the python package setup, works for me!

PR #249, opened 2023-05-07
PEP 561 typed package

### LINKED ISSUE #248: PEP 561 compatibility: py.typed marker?
Importing `edg` in my editor with the pylsp-mypy language server, I get a warning that there is no `py.typed` marker (or stubs).

I'm not terribly familiar with python, but it sure looks like you all are using type annotations. If I'm reading [this](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages) right all that'd be needed is adding an empty `py.typed` file to the packages (and maybe something to indicate that it should be added to the packaged module?).

imuli:
resolves #248

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,9 @@
 include edg/edg-compiler-precompiled.jar
+include edg_core/py.typed
+include edg_hdl_server/py.typed
+include edgir/py.typed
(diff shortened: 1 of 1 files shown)


ducky64 (approved):
Looks good, thanks for catching this!

PR #314, opened 2024-02-28
Pcbbot

Suke0811:
Robot driver that uses a ESP32 w/ camera and has student-proofing
  Key features:
  - USB-C receptacle for power input and battery charging.
  - LiPo battery connector with voltage regulation and protection circuits.
  - Power management with priority power selection, fuse protection, and gate control.
  - 3.3V voltage regulation for the main logic level power supply.
  - Integrated battery charging circuit with status indication.
  - I2C communication interface with pull-up resistors and test points.
  - Time-of-flight (ToF) sensor array for distance measurement.
  - Inertial Measurement Unit (IMU) and magnetometer for orientation sensing.
  - IO expander for additional GPIOs and a thru-hole RGB LED indicator.
  - OLED display
  - PWM Servo motor
  - Neopixel LED array for RGB lighting
  - Camera module with voltage regulation for image capture.
  - Digital switch for power control.
  - Keyboard mechanical switch with RGB LED
  - Ducky touch button
  
  Known issues:
  - Charging ic is not reverse protected
  - MCU does not get turned off with the gate when powered by the USB and battery, though the vbatt line turns off. (by design)


ducky64 (changes requested):
HDL looks good.

Can you update the .kicad_pcb file to the latest as-sent-to-fab, and also add the netlist to git (by running the entire unit test suite and committing the generated `.net` file)? 

It also looks like there's a mypy error. `connected_from` doesn't understand Connection types, only Ports, so just change it from `self.vbatt` (a Connection) to `self.gate.pwr_out` (the Port that is in the connection).

I should fix that at the root, but the fix isn't straightforward - so let's work around it and get this merged sooner.

Suke0811:
@ducky64 
.kicad_pcb and .net for Pcbbot have been updated. 

> It also looks like there's a mypy error. connected_from doesn't understand Connection types, only Ports, so just change it from self.vbatt (a Connection) to self.gate.pwr_out (the Port that is in the connection).

With this fix 7a0709a now all the tests pass.

ducky64 (approved):
lgtm!

PR #318, opened 2024-03-12
Add:  opa189 as one of the parts option

Suke0811:
Adding Opa189 for OPA197 since OPA197 is out of stock.


Suke0811:
What is the best way to do this? @ducky64

ducky64:
The right way to do this (which I'll probably get to in a week or so for an iteration of the SMU board) is to model the part, kinda copy-paste the OPA197 but update the specs for OPA189. If you want to do this before I do, feel free to submit it as its own PR.

Since OPA197 is pin compatible and close-enough for your application, you can also use a top-level refinement to override both the part number and lcsc_part. Not super elegant, but doesn't need any infrastructure / library changes.

PR #329, opened 2024-04-02
Ballu

Suke0811:
- Update: FetPowerSwitch with btn_in
- Add: The BALLU PCB board uses FetPowerSwitch with a directional switch center as the power button.

- [ ] TODO: manufacture


ducky64 (changes requested):
I see this adds an example and adds the soft power button to the core library (instead of just in the multimeter example, as it was before). For that, can you:
- deduplicate with the multimeter example, and refactor the multimeter example to use the core library
- update comments on the schematic about how it's supposed to be used
- add a description into the class docstring describing how it's supposed to be used, and that users should be careful not to connect the button to any downstream circuit, and that it's not checked
- maybe a better name - SoftPowerGate? I think these are typically called soft power switches, but this subcircuit has the button / switch external.
- would be preferable (but not required) if the library change was separate from the example

Will look over the example circuit hdl and layout momentarily.

ducky64 (commented):
I skimmed the HDL and layout and it looks reasonable for fabrication. Hopefully it works!

One suggestion is to silkscreen label the connectors, so it's obvious which is for what.

Were you also going to try to add the reverse protected charger circuit? I think as of now nothing is reverse protected?

You could also consider using a copper pour for the inner power layer, then have higher priority pours to make sub-sections (like for battery voltage, if your main pour is 3v3).

However, if you want to merge the full example into the main codebase, we need to have a longer discussion on style and form - the examples should be examples of best practices, style, and structure. In any case, I think it makes sense to merge the FetPowerGate (see prior comment) as a standalone library improvement.

As an alternative to merging this example, it might be possible to create your own repository and submodule PolymorphicBlocks - allowing you to keep your own repository of convenience Blocks that might not be as generalizable as the current library blocks. There might need to be some IDE changes to make this work, we can also discuss that.

Suke0811:
@ducky64 Thanks for the review!
1. I'll put this into fabrication tonight after adding labels, so fingers crossed!
2. Yeah, I agree, I can pull out the FetGateSwitch/SoftGateSwitch to an independent PR
3. Comments regarding SoftGateSwitch will be applied to that branch (sometime this week)
4. The main test_ballu doesn't need to be merged. It is very close to PCBbot but with a directional soft power switch
5. There are no reverse protections on the charger, so reversed lipo will blow the charging ic 

We plan to make many more of this board with more tweaks, so the reverse protection will be added later...

PR #334, opened 2024-04-05
SoftwarePowerGate

Suke0811:
Add: SoftPowerGate and switch

Originally from #329


electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,107 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+                        )
+                      })
+
+class SoftPowerSwitch(SoftPowerGate):
+  """A soft power switch that inherits from SoftPowerGate, with an additional
+  button for manual power control.
+  """
+  def contents(self):
+    self.btn = self.Block(Switch(voltage=0*Volt(tol=0)))  # TODO - actually model switch voltage
+    super().contents()
```
Suke0811, line 369: This inevitably causes an error since the btn_in is no longer connected or needed here. 

- [ ] Q is there a way to remove the port in the parent?

electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,107 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+    self.control = self.Port(DigitalSink.empty())  # digital level control - gnd-referenced NFET gate
+
+    self.pull_resistance = self.ArgParameter(pull_resistance)
+    self.amp_resistance = self.ArgParameter(amp_resistance)
+    self.diode_drop = self.ArgParameter(diode_drop)
+
+  def contents(self):
+    super().contents()
+
+    control_voltage = self.control.link().voltage.hull(RangeExpr.ZERO,)
```
ducky64, line 303: If this is the max voltage across ctl_diode and btn_diode, it may be more correct to use the voltage on the btn_in port

electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,107 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+    ))
+
+    self.btn_diode = self.Block(Diode(
+      reverse_voltage=control_voltage,
+      current=RangeExpr.ZERO,  # effectively no current
+      voltage_drop=self.diode_drop,
+      reverse_recovery_time=RangeExpr.ALL
+    ))
+
+    self.import_kicad(self.file_path("resources", f"{self.__class__.__name__}.kicad_sch"),
```
ducky64, line 339: I think the schematic still has the switch in it? Does that need to get removed?

electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,107 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+  on by an external signal, and provides the button output as a signal.
+
+  Attributes:
+    pull_resistance: The resistance of the pull-up resistor.
+    amp_resistance: The resistance of the amplifier resistor.
+    diode_drop: The voltage drop across the diode.
+
+  Ports:
+    pwr_in: A voltage sink port for the input power.
+    pwr_out: A voltage source port for the output power.
```
ducky64, line 279:
```suggestion
    pwr_out: A voltage source port for the output power, gated by the power button.
```

electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,107 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+  """A high-side PFET power gate that has a button to power on, can be latched
+  on by an external signal, and provides the button output as a signal.
+
+  Attributes:
+    pull_resistance: The resistance of the pull-up resistor.
+    amp_resistance: The resistance of the amplifier resistor.
+    diode_drop: The voltage drop across the diode.
+
+  Ports:
+    pwr_in: A voltage sink port for the input power.
```
ducky64, line 278: Hm, stylistically I might omit the port type information (since it's redundant with the code). Or if you think the type information is important, maybe use the port type name directly, eg `VoltageSink`?

electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,107 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+    pwr_out: A voltage source port for the output power.
+    gnd: ground.
+    btn_out: A digital single source port for the button output signal.
```
ducky64, line 281:
```suggestion
    btn_out: An optional digital single source port for the button output signal, allowing the button state to be read independently of the control signal.
```

ducky64 (changes requested):
I think the overall structure is good, though we could work on the details.

I think the same PR should also refactor the multimeter example to use this shared library to deduplicate code - it should just be removing the code from the multimeter file and changing the design to use this one.

electronics_lib/PowerConditioning.py
Suke0811: I think btn_in/btn_out/control are better to have descriptions, and I have other ports for the sake of completeness. Then, update to use the type name directly like suggested?

Suke0811: Also this is google styled docsting but if you have other preference i can change it

electronics_lib/PowerConditioning.py
Suke0811: updated

electronics_lib/PowerConditioning.py
Suke0811: I think it should be removed already

image

electronics_lib/PowerConditioning.py
Suke0811: >I'm on the fence as to whether a convenience wrapper for the SoftPowerGate + switch makes sense, since it does so little extra. But on the positive side, it's more error-proof in the common case, while advanced users can use custom switches like encoders - and I think it'd be worth it? Thoughts?

I think it is good to add SoftPowerGate + Switch  If people anyway need to add a button at least always.
 In most cases, people would do refinement on the switch to swap up the button and it should be enough I assume?
 
 > switches like encoders
 Spin to turn on 🙃

Suke0811: Composition vs Inheritance and use composition for this

Suke0811: '    control_voltage = self.control.link().voltage.hull(ground)'

electronics_lib/PowerConditioning.py
Suke0811: resolved with composition ver 
18eef59

ducky64:
Try making `btn_in` a DigitalBidir - which is more accurate from a modeling perspective, since it's both an output (in that there's a pullup) and an input (in that it reads the button state). DigitalBidir should be able to indicate the output is pullup capable.

electronics_lib/PowerConditioning.py
```diff
@@ -263,3 +263,112 @@ def connected_from(self, gnd: Optional[Port[VoltageLink]] = None, pwr_hi: Option
...
+
+class SoftPowerSwitch(PowerSwitch, Block):
+  """A software power switch that adds a power button a user can turn on
+  """
+  @init_in_parent
+  def __init__(self, pull_resistance: RangeLike = 10 * kOhm(tol=0.05), amp_resistance: RangeLike = 10 * kOhm(tol=0.05),
+               diode_drop: RangeLike = (0, 0.4) * Volt):
+    super().__init__()
+
+    self.gnd = self.Port(Ground.empty(), [Common])
```
ducky64, line 360: Stylistically, consider instantiating the SoftPowerGate and exporting its gnd. Then, create and connect btn later - can be in `contents`

ducky64 (commented):
Overall looks good with one stylistic comment and pending the fix of the unit test by using DigitalBidir.

Also, make sure you run the unit test suite locally and commit the updated netlist.

electronics_lib/PowerConditioning.py
Suke0811: applied in 9c3f75d

ducky64 (approved):
lgtm! merge at your discretion!

PR #343, opened 2024-04-23
Experimental API for port and param docs

ducky64:
Adds an optional `doc='...'` argument to `self.Port(...)`, `self.Parameter(...)` and `self.ArgParameter(...)`. Docs are stored in the block, similar to how optional and port tags are tracked. The idea is this keeps documentation close to the port definition and (in the future) allows tools that are port-/parameter-aware to generate documentation specific to the data model here.

Python doesn't have a accepted convention for class attribute docs, so we're going to make our own. [PEP 224](https://peps.python.org/pep-0224/) was a proposal for attribute docstrings, but was rejected.

Quick informal conventions:
- Type information should not be repeated in the docstring.
- Other arguments to Port should not be repeated in the docstring (e.g., optional, port tags). Tools should automatically add these when generating visualizations / documentation.
- Documentation should not simply repeat the port name, but rather add more information / context - `pwr`, `gnd` would not need docstrings unless the docstrings add information, e.g. disambiguate between multiple power rails or `pwr_in` vs `pwr_out` or highlight odd behavior.

### DIFF
--- a/edg_core/Blocks.py
+++ b/edg_core/Blocks.py
@@ -244,8 +244,12 @@ class BaseBlock(HasMetadata, Generic[BaseBlockEdgirType]):
 
     # TODO delete type ignore after https://github.com/python/mypy/issues/5374
     self._parameters: SubElementDict[ConstraintExpr] = self.manager.new_dict(ConstraintExpr)  # type: ignore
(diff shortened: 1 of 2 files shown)


Suke0811 (approved):
Good

PR #363, opened 2024-07-10
fix Range.fuzzy_in() typo bug and add 1210 capacitor smt size

lrobot:
some mini improvement when debug the the capacitor generator:

I found most 47uF capacitor is in big smt size 1210 that is current ignored in JlcCapacitor.py csv part loader.

- fix Range.fuzzy_in() typo bug and add 1210 capacitor smt size

### DIFF
--- a/edg/parts/JlcCapacitor.py
+++ b/edg/parts/JlcCapacitor.py
@@ -13,12 +13,14 @@ class JlcCapacitor(TableDeratingCapacitor, CeramicCapacitor, SmdStandardPackageS
     '0603': 'Capacitor_SMD:C_0603_1608Metric',
     '0805': 'Capacitor_SMD:C_0805_2012Metric',
     '1206': 'Capacitor_SMD:C_1206_3216Metric',
+    '1210': 'Capacitor_SMD:C_1210_3225Metric',
     '1812': 'Capacitor_SMD:C_1812_4532Metric',
 
     'C0402': 'Capacitor_SMD:C_0402_1005Metric',
     'C0603': 'Capacitor_SMD:C_0603_1608Metric',
     'C0805': 'Capacitor_SMD:C_0805_2012Metric',
     'C1206': 'Capacitor_SMD:C_1206_3216Metric',
+    'C1210': 'Capacitor_SMD:C_1210_3225Metric',
     'C1812': 'Capacitor_SMD:C_1812_4532Metric',
   }
   DERATE_VOLTCO_MAP = {  # in terms of %capacitance / V over 3.6
@@ -26,6 +28,7 @@ class JlcCapacitor(TableDeratingCapacitor, CeramicCapacitor, SmdStandardPackageS
     'Capacitor_SMD:C_0603_1608Metric': float('inf'),  # not supported, should not generate below 1uF
     'Capacitor_SMD:C_0805_2012Metric': 0.08,
     'Capacitor_SMD:C_1206_3216Metric': 0.04,
+    'Capacitor_SMD:C_1210_3225Metric': 0.04,
     'Capacitor_SMD:C_1812_4532Metric': 0.04,  # arbitrary, copy from 1206
   }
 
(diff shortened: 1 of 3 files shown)


edg/parts/JlcCapacitor.py
```diff
@@ -8,24 +8,28 @@
...
-
+    '402': 'Capacitor_SMD:C_0402_1005Metric',
     '0402': 'Capacitor_SMD:C_0402_1005Metric',
+    '603': 'Capacitor_SMD:C_0603_1608Metric',
     '0603': 'Capacitor_SMD:C_0603_1608Metric',
+    '805': 'Capacitor_SMD:C_0805_2012Metric',
     '0805': 'Capacitor_SMD:C_0805_2012Metric',
     '1206': 'Capacitor_SMD:C_1206_3216Metric',
+    '1210': 'Capacitor_SMD:C_1210_3225Metric',
     '1812': 'Capacitor_SMD:C_1812_4532Metric',
```
ducky64, line 19: Support for 1210 makes sense.

I'm less convinced about the need to add parsing for `402`/`603`/`805`, I looked through the pruned table and only `603` shows up for capacitors, in one line item.

In the future, the right answer might be to delegate parsing / name canonicalization to another project, like https://github.com/yaqwsx/jlcparts. The current pruned parts library is pretty outdated, though it still does work, and has the benefit of repeatable unit tests.

lrobot: I found about 1106 capactor with package miss "0" header.
0F2DA278-85F9-46D9-ABCB-625022C8BB78
leave this may compatible process may better for library user?

ducky64: I think that's Excel interpreting the field as a number and dropping the leading zero in its presentation. In reality most of those entries (with only 3-4 exceptions, I think) have the leading zero.

In my version of Excel there's the option to do data import without interpreting the data types, that should preserve the original formatting.

lrobot: Ok. Got it

edg/parts/JlcCapacitor.py
```diff
@@ -38,10 +42,10 @@ def __init__(self, *args, capacitance_minimum_size: BoolLike = True, **kwargs):
...
   def _make_table(cls) -> PartsTable:
     CAPACITOR_MATCHES = {
-      'nominal_capacitance': re.compile("(^|\s)([^±]\S+F)($|\s)"),
-      'tolerance': re.compile("(^|\s)(±\S+[%F])($|\s)"),
-      'voltage': re.compile("(^|\s)(\d\S*V)($|\s)"),  # make sure not to catch 'Y5V'
-      'tempco': re.compile("(^|\s)([CXYZ]\d[GPRSTUV])($|\s)"),
+      'nominal_capacitance': re.compile("(^|\\s)([^±]\\S+F)($|\\s)"),
+      'tolerance': re.compile("(^|\\s)(±\\S+[%F])($|\\s)"),
+      'voltage': re.compile("(^|\\s)(\\d\\S*V)($|\\s)"),  # make sure not to catch 'Y5V'
+      'tempco': re.compile("(^|\\s)([CXYZ]\\d[GPRSTUV])($|\\s)"),
```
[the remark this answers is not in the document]
lrobot: yes. method r"" is better

edg/parts/JlcCapacitor.py
```diff
@@ -72,7 +77,9 @@ def parse_row(row: PartsTableRow) -> Optional[Dict[PartsTableColumn, Any]]:
         new_cols.update(cls._parse_jlcpcb_common(row))
 
         return new_cols
-      except (KeyError, PartParserUtil.ParseError):
+      except (KeyError, PartParserUtil.ParseError) as e:
+        # when some JLC part didnot loaded correctly, can use print to find cause
+        # print('KeyError', row[cls._PACKAGE_HEADER], row.value, e)
```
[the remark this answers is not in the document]
lrobot: Just remove this deadcode,

edg/parts/JlcCapacitor.py
```diff
@@ -54,9 +57,10 @@ def parse_row(row: PartsTableRow) -> Optional[Dict[PartsTableColumn, Any]]:
         footprint = cls.PACKAGE_FOOTPRINT_MAP[row[cls._PACKAGE_HEADER]]
         extracted_values = cls.parse(row[cls.DESCRIPTION_COL], CAPACITOR_MATCHES)
 
-        tempco = extracted_values['tempco'][1]
-        if tempco[0] not in ('X', 'C') or tempco[2] not in ('R', 'S', 'G'):
-          return None
+        # some JLC part have no tempco info, but still want use it
+        # tempco = extracted_values['tempco'][1]
+        # if tempco[0] not in ('X', 'C') or tempco[2] not in ('R', 'S', 'G'):
+        #   return None
```
ducky64, line 63: I don't think this filtering step should be removed. Tempco doesn't just affect temperature stability and tolerances, I believe it's also correlated with capacitance derating - where at DC load, the capacitor has less capacitance. Worse capacitors (eg, Y5V) typically have worse derating (and are less 'ideal') than better ones (eg, X5R), though the benefit is that they have higher capacitance.

A future PR might address this by adding a parameter `CeramicCapacitor.minimum_tempco`, which would default to `X5R` (following the principle of erring on the side of overkill rather than wrong) but would still allow the user (via refinements) to specify an alternative.

ducky64 (changes requested):
Other changes are good, but somehow I missed a comment about temperature coefficient from the last round - sorry!

I distinctly remember typing it up, perhaps I forgot to commit the comment? But should be a pretty straightforward change.

edg/parts/JlcCapacitor.py
lrobot: Ok, So this PR will leave this untouched.  Thanks for give your deep consider

ducky64 (approved):
Thanks! Looks good!

PR #369, opened 2024-07-24
Disallow inner connects

ducky64:
Adds a frontend error so it doesn't result in a confusing compiler internal error.

Also fixes an unrelated type annotation issue flagged by a new version of mypy.

### DIFF
--- a/edg/core/Blocks.py
+++ b/edg/core/Blocks.py
@@ -219,7 +219,7 @@ class DescriptionString():
       self.ref = ref
       self.units = units
 
-    def set_elt_proto(self, pb, ref_map):
+    def set_elt_proto(self, pb, ref_map=None):
       new_phrase = pb.description.add()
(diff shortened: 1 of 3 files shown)


Suke0811:
With this 
```
self.connect(self.conn.miso, self.mag.spi.miso)
```
The attempt to connect inner ports errors outs properly as follows
```
PolymorphicBlocks.edg.core.HdlUserExceptions.UnconnectableError: Inaccessible port: cannot connect inner ports on inner blocks
```

Suke0811 (approved):
With this 
```
self.connect(self.conn.miso, self.mag.spi.miso)
```
The attempt to connect inner ports errors outs properly as follows
```
PolymorphicBlocks.edg.core.HdlUserExceptions.UnconnectableError: Inaccessible port: cannot connect inner ports on inner blocks
```

PR #377, opened 2024-09-20
Update: Camera_Ov2640_Fpc24 to use non strict compatible

Suke0811:
Update: modifying the camera code to accept 3v3 if nonstirct is true. 
The code is used in pcbbot and updated accordingly

### DIFF
--- a/edg/parts/Camera_Ov2640_Fpc24.py
+++ b/edg/parts/Camera_Ov2640_Fpc24.py
@@ -13,7 +13,7 @@ class Ov2640_Fpc24_Device(InternalSubcircuit, Nonstrict3v3Compatible, Block):
         self.dovdd = self.Export(self.conn.pins.request('14').adapt_to(VoltageSink(
             voltage_limits=self.nonstrict_3v3_compatible.then_else(
                 (1.71, 4.5)*Volt,  # Table 6, absolute maximum (Table 5) is 4.5v
-                (1.71, 3.3)*Volt,),
+                (1.71, 3.3)*Volt),
             current_draw=(6, 15)*mAmp  # active typ to max
         )))
         self.dvdd = self.Export(self.conn.pins.request('15').adapt_to(VoltageSink(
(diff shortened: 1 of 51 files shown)


examples/test_pcbbot.py
```diff
@@ -220,7 +220,6 @@ def refinements(self) -> Refinements:
         (CompactKeystone5015, ['lcsc_part'], 'C5199798'),  # RH-5015, which is actually in stock
 
         # the camera recommended specs are excessively tight, so loosen them a bit
-        (Ov2640_Fpc24, ['device', 'dovdd', 'voltage_limits'], Range(1.71, 4.5)),
         (Ov2640_Fpc24, ['device', 'dvdd', 'voltage_limits'], Range(1.1, 1.36)),  # allow 1v2
```
ducky64, line 223: If this is deleted will it still work? Since this change also fixes the 1.2v limit

edg/parts/Camera_Ov2640_Fpc24.py
```diff
@@ -11,11 +11,13 @@ def __init__(self) -> None:
         self.dgnd = self.Export(self.conn.pins.request('10').adapt_to(Ground()))
         self.agnd = self.Export(self.conn.pins.request('23').adapt_to(Ground()))
         self.dovdd = self.Export(self.conn.pins.request('14').adapt_to(VoltageSink(
-            voltage_limits=(1.71, 3.3)*Volt,  # Table 6, absolute maximum (Table 5) is 4.5v
+            voltage_limits=self.nonstrict_3v3_compatible.then_else(
+                (1.71, 4.5)*Volt,  # Table 6, absolute maximum (Table 5) is 4.5v
+                (1.71, 3.3)*Volt,),
```
ducky64, line 16:
```suggestion
                (1.71, 3.3)*Volt),
```
stray trailing comma

ducky64 (approved):
Looks good, one style comment inline.

examples/test_pcbbot.py
Suke0811: yeah removing `  (Ov2640_Fpc24, ['device', 'dvdd', 'voltage_limits'], Range(1.1, 1.36)),  # allow 1v2` still works with no error. pushed with f0170fe1cef1b1cab850783d722505e51ea1171f

ducky64 (approved):

PR #379, opened 2024-09-20
Update: handling zero voltage rating cases

Suke0811:
Do we still need to merge this?


ducky64 (approved):
Yes! This is still useful, it allows resistors without a voltage rating to be selected.

PR #381, opened 2024-09-20
Add: Ina219 current/voltage sensor over I2C

Suke0811:
- [x] Hardware to be tested, at least it seems to be alive so far

### DIFF
--- a/edg/parts/Ina219.py
+++ b/edg/parts/Ina219.py
@@ -28,8 +28,8 @@ class Ina219_Device(InternalSubcircuit, JlcPart, FootprintBlock, Block):
         self.i2c.sda.init_from(dio_sda_model)
         self.i2c.scl.init_from(DigitalSink.from_bidir(dio_model))
 
-        self.in_pos = self.Port(AnalogSink(voltage_limits=(0.0, 26) * Volt))
-        self.in_neg = self.Port(AnalogSink(voltage_limits=(0.0, 26) * Volt))
+        self.in_pos = self.Port(AnalogSink(voltage_limits=(-0.3, 26) * Volt))
+        self.in_neg = self.Port(AnalogSink(voltage_limits=(-0.3, 26) * Volt))
 
     def contents(self):
         super().contents()
--- a/edg/parts/Ina219.py
+++ b/edg/parts/Ina219.py
@@ -49,6 +49,7 @@ class Ina219_Device(InternalSubcircuit, JlcPart, FootprintBlock, Block):
             datasheet='https://www.ti.com/lit/ds/symlink/ina219.pdf'
         )
         self.assign(self.lcsc_part, "C87469")
+        self.assign(self.actual_basic_part, False)
 
 
 class Ina219(CurrentSensor, Block):
--- a/edg/parts/Ina219.py
+++ b/edg/parts/Ina219.py
@@ -26,7 +26,7 @@ class Ina219_Device(InternalSubcircuit, JlcPart, FootprintBlock, Block):
 
         self.i2c = self.Port(I2cTarget(DigitalBidir.empty(), addresses=[0x40]))
         self.i2c.sda.init_from(dio_sda_model)
-        self.i2c.scl.init_from(dio_model)
+        self.i2c.scl.init_from(DigitalSink.from_bidir(dio_model))
 
         self.in_pos = self.Port(AnalogSink(voltage_limits=(0.0, 26) * Volt))
         self.in_neg = self.Port(AnalogSink(voltage_limits=(0.0, 26) * Volt))
--- a/edg/parts/Ina219.py
+++ b/edg/parts/Ina219.py
@@ -13,20 +13,17 @@ class Ina219_Device(InternalSubcircuit, JlcPart, FootprintBlock, Block):
         ))
         self.gnd = self.Port(Ground())
 
-        dio_model = DigitalBidir.from_supply(
+        self.i2c = self.Port(I2cTarget(DigitalBidir.empty(), addresses=[0x40]))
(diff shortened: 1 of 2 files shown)


Suke0811:
Q: 
any idea why this is the case?
`edg/parts/Ina219.py:29: error: Argument 1 to "init_from" of "Port" has incompatible type "DigitalBidir"; expected "DigitalSink"  [arg-type]`
When i locally run mypy, i got the same error but otherwise, i don't see the error.

edg/parts/Ina219.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        dio_sda_model = DigitalBidir.from_supply(
+            self.gnd, self.vs,
+            voltage_limit_abs=(-0.3 * Volt, 6.0 * Volt),
+            input_threshold_factor=(0.3, 0.7)
+        )
+
+        self.i2c = self.Port(I2cTarget(DigitalBidir.empty(), addresses=[0x40]))
+        self.i2c.sda.init_from(dio_sda_model)
+        self.i2c.scl.init_from(dio_model)
```
ducky64, line 29: I think this might be the culprit, i2c.scl needs to be init_from DigitalSink since the type checker here is strict.

Does this work?
```suggestion
        self.i2c.scl.init_from(DigitalSink.from_bidir(dio_model))
```

Suke0811: Yeah, this works, and I removed the mypy error.  Ina219 compiles as before

edg/parts/Ina219.py
```diff
@@ -0,0 +1,77 @@
...
+                '4': self.vs,
+                '5': self.i2c.scl,
+                '6': self.i2c.sda,
+                '7': self.gnd,  # TODO: make this address selectable, sa0 pin
+                '8': self.gnd  # TODO: make this address selectable, sa1 pin
+            },
+            mfr='Texas Instruments', part='INA219AIDCNR',
+            datasheet='https://www.ti.com/lit/ds/symlink/ina219.pdf'
+        )
+        self.assign(self.lcsc_part, "C87469")
```
ducky64, line 51:
```suggestion
        self.assign(self.lcsc_part, "C87469")
        self.assign(self.actual_basic_part, False)
```
Should also add that it's not a basic part instead of leaving it unassigned

edg/parts/Ina219.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        dio_sda_model = DigitalBidir.from_supply(
+            self.gnd, self.vs,
+            voltage_limit_abs=(-0.3 * Volt, 6.0 * Volt),
+            input_threshold_factor=(0.3, 0.7)
+        )
+
+        self.i2c = self.Port(I2cTarget(DigitalBidir.empty(), addresses=[0x40]))
+        self.i2c.sda.init_from(dio_sda_model)
+        self.i2c.scl.init_from(DigitalSink.from_bidir(dio_model))
```
ducky64, line 29:
```suggestion
        self.i2c = self.Port(I2cTarget(DigitalBidir.empty(), addresses=[0x40]))
        self.i2c.sda.init_from(DigitalBidir.from_supply(
            self.gnd, self.vs,
            voltage_limit_abs=(-0.3, 6.0) * Volt,
            input_threshold_factor=(0.3, 0.7)
        ))
        self.i2c.scl.init_from(DigitalSink.from_supply(
            self.gnd, self.vs,
            voltage_limit_tolerance=(-0.3, 0.3) * Volt,
            input_threshold_factor=(0.3, 0.7)
        ))
```

Consider just inlining both models, since they're not shared with anything else.

It would make sense to have a shared model on IN+/IN- if you'd like, though.

edg/parts/Ina219.py
```diff
@@ -0,0 +1,77 @@
...
+
+        self.in_pos = self.Port(AnalogSink(voltage_limits=(0.0, 26) * Volt))
+        self.in_neg = self.Port(AnalogSink(voltage_limits=(0.0, 26) * Volt))
```
ducky64, line 32:
```suggestion
        self.in_pos = self.Port(AnalogSink(voltage_limits=(-0.3, 26) * Volt))
        self.in_neg = self.Port(AnalogSink(voltage_limits=(-0.3, 26) * Volt))
```
Looks like the analog input rating is down to -0.3v

ducky64 (changes requested):
Mostly looks good, few fixes inline.
Let's get it tested before merging?

edg/parts/Ina219.py
Suke0811: added

Suke0811:
image
The sensor seems to be alive and the voltage is reasonable. 

I used https://github.com/RobTillaart/INA219 instead of the adafruit library since they assume a predefined shunt resistor and no mechanisms to change it without modifying the library.

ducky64 (approved):
hooray!

ducky64:
Actually, would it make sense to specify the default shunt resistance t 0.1ohm for compatibility with the Adafruit board?

Suke0811:
Adafruit lib has hardcoded calibration with a set shunt resistor, so maybe it's not good for general purposes. In the inline comment, they say implement your own lol
```arduino
  void setCalibration_32V_2A();
  void setCalibration_32V_1A();
  void setCalibration_16V_400mA();
  ```
  
  ```
    // By default we use a pretty huge range for the input voltage,
  // which probably isn't the most appropriate choice for system
  // that don't use a lot of power.  But all of the calculations
  // are shown below if you want to change the settings.  You will
  // also need to change any relevant register settings, such as
  // setting the VBUS_MAX to 16V instead of 32V, etc.
  ```

ducky64:
Fair enough! 2mOhm it is, and the datasheet is probably a more authoritative source anyways.

PR #382, opened 2024-09-20
Add: can diff side test points

Suke0811:
Adding the testpoints for can diff side. The controller side is already in the lib.


ducky64 (changes requested):
Block itself looks good. Though I think it also needs to be added to __init__.py?

Suke0811:
> Block itself looks good. Though I think it also needs to be added to **init**.py?

That's true, and I added it in init with [3b3a4fb](https://github.com/BerkeleyHCI/PolymorphicBlocks/pull/382/commits/3b3a4fb89c7a036724798cad38fbf137456e1ddb)

PR #383, opened 2024-09-20
Add: Tle5012b encoder

Suke0811:
A 15 bit magnetic encoder parts available on jlcpcb

### DIFF
--- a/edg/parts/Tle5012b.py
+++ b/edg/parts/Tle5012b.py
@@ -1,0 +1,77 @@
+from edg.abstract_parts import *
+from edg.parts.JlcPart import JlcPart
+
+
+class Tle5012b_Device(InternalSubcircuit, FootprintBlock, JlcPart):
+    def __init__(self) -> None:
+        super().__init__()
+        self.vdd = self.Port(VoltageSink(
+            voltage_limits=(3.0, 5.5) * Volt,
+            current_draw=(5 * uAmp, 16 * mAmp)
+        ))
+        self.gnd = self.Port(Ground())
+
+        dio_model = DigitalBidir.from_supply(
+            self.gnd, self.vdd,
+            voltage_limit_tolerance=(-0.3, 0.3) * Volt,
+            input_threshold_factor=(0.3, 0.7),
+            current_limits=(5 * uAmp, 25 * mAmp)
+        )
+        dio_model_sck = DigitalSink.from_supply(self.gnd, self.vdd,
+                                                voltage_limit_tolerance=(-0.3, 0.3) * Volt)
+        self.sck = self.Port(dio_model_sck)
+        self.csq = self.Port(dio_model_sck)
+        self.data = self.Port(DigitalSink.from_bidir(dio_model))
+
+        if_dio_model = DigitalBidir.from_supply(
+            self.gnd, self.vdd,
+            voltage_limit_tolerance=(-0.3, 0.3) * Volt,
+            input_threshold_factor=(0.3, 0.7),
+            current_limits=(5 * uAmp, 15 * mAmp)
+        )
+
+        self.ifa = self.Port(DigitalSource.from_bidir(if_dio_model), optional=True)
+        self.ifb = self.Port(DigitalSource.from_bidir(if_dio_model), optional=True)
+        self.ifc = self.Port(DigitalSource.from_bidir(if_dio_model), optional=True)
+
+    def contents(self) -> None:
+        self.footprint(
+            'U', 'fp:SO8_PG-DSO-8_INF',
+            {
+                '1': self.ifc,
+                '2': self.sck,
+                '3': self.csq,
+                '4': self.data,
+                '5': self.ifa,
+                '6': self.vdd,
+                '7': self.gnd,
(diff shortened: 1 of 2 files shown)


edg/parts/Tle5012b.py
```diff
@@ -0,0 +1,77 @@
...
+            },
+            mfr='Infineon', part='Tle5012b',
+            datasheet='https://www.infineon.com/cms/en/product/sensor/magnetic-sensors/magnetic-position-sensors/angle-sensors/tle5012b-e1000/'
+        )
+        self.assign(self.lcsc_part, 'C123083')
+        self.assign(self.actual_basic_part, False)
+
+
+class Tle5012b(Magnetometer, DefaultExportBlock):
+    """Angle sensor based on Giant Magneto Resistance (GMR) for precise angular position sensing."""
```
ducky64, line 58: Consider quantifying the precision. I think it's a 15-bit angle output?

edg/parts/Tle5012b.py
```diff
@@ -0,0 +1,77 @@
...
+        self.pwr = self.Export(self.ic.vdd, [Power])
+
+        self.sck = self.Export(self.ic.sck, optional=True)
+        self.csq = self.Export(self.ic.csq, optional=True)
+        self.data = self.Export(self.ic.data, optional=True)
+
+        self.ifa = self.Export(self.ic.ifa, optional=True,
+                               doc="IIF Phase A / Hall Switch Signal 1 / PWM / SPC output (input for SPC trigger only)")
+        self.ifb = self.Export(self.ic.ifb, optional=True, doc="IIF Phase B / Hall Switch Signal 2")
+        self.ifc = self.Export(self.ic.ifc, optional=True, doc="External Clock / IIF Index / Hall Switch Signal 3")
```
ducky64, line 73: Do you use these? If not, I'd consider leaving these out for simplicity, for now.

Part of the reason is that design-side, I'd like the application circuit to provide high-level interfaces, and the application circuit could figure out how to map like PWM to device pins and enforce constraints (such as only one of the IO options can be active)

edg/parts/Tle5012b.py
```diff
@@ -0,0 +1,77 @@
...
+            current_limits=(5 * uAmp, 15 * mAmp)
+        )
+
+        self.ifa = self.Port(DigitalSource.from_bidir(if_dio_model), optional=True)
+        self.ifb = self.Port(DigitalSource.from_bidir(if_dio_model), optional=True)
+        self.ifc = self.Port(DigitalSource.from_bidir(if_dio_model), optional=True)
+
+    def contents(self) -> None:
+        self.footprint(
+            'U', 'fp:SO8_PG-DSO-8_INF',
```
ducky64, line 39:
```suggestion
            'U', 'Package_SO:SOP-8_3.9x4.9mm_P1.27mm',
```
Can a standard KiCad footprint be used? This one should be compatible, though the pads stretch more on the inside

edg/parts/Tle5012b.py
```diff
@@ -0,0 +1,77 @@
...
+            voltage_limits=(3.0, 5.5) * Volt,
+            current_draw=(5 * uAmp, 16 * mAmp)
+        ))
+        self.gnd = self.Port(Ground())
+
+        dio_model = DigitalBidir.from_supply(
+            self.gnd, self.vdd,
+            voltage_limit_tolerance=(-0.3, 0.3) * Volt,
+            input_threshold_factor=(0.3, 0.7),
+            current_limits=(5 * uAmp, 25 * mAmp)
```
ducky64, line 18: Where'd this come from?
It seems that it can sink up to -25mA and has a pull-up rating of 3uA? I don't see a source current limit even though there's a section on the SSC interface in push-pull configuration.

edg/parts/Tle5012b.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        dio_model_sck = DigitalSink.from_supply(self.gnd, self.vdd,
+                                                voltage_limit_tolerance=(-0.3, 0.3) * Volt)
+        self.sck = self.Port(dio_model_sck)
+        self.csq = self.Port(dio_model_sck)
```
ducky64, line 23:
```suggestion
        self.sck = self.Port(DigitalSink.from_bidir(dio_model))
        self.csq = self.Port(DigitalSink.from_bidir(dio_model))
```
I think most of the pindefs can be reused?

ducky64 (commented):
Mostly looks ok, some suggestions. Consider eliminating the IIF pins if you're not using them right now, raw pins is lower-level of abstraction than I'd like if we don't need them yet.

This device has a weird interface, I haven't seen SSC elsewhere. How are you connecting this to the ESP32? Is there a way this should be adapted to a more standard interface?

Suke0811:
This device uses stm32 instead of esp32, which is why it uses SSC. Are you still interested in merging this?

ducky64:
Hm. I think if it can't present one of the existing high-level digital interface busses (like SPI), it might not fit with the level-of-abstraction of the rest of the library.

Perhaps eventually we'll support SSC, but it doesn't seem common enough right now.

PR #396, opened 2025-03-18
Add: Lsm6dsv16x high end smartphone IMU with qvar support.

Suke0811:
This adds Lsm6dsv16x high end smartphone IMU. 
Very similar and electrically compatible with [Imu_Lsm6ds3trc](https://github.com/BerkeleyHCI/PolymorphicBlocks/blob/561da36d69b866f58d5fc6faa8a48c919b618e36/edg/parts/Imu_Lsm6ds3trc.py), but this chip has extended voltage range and lower power consumptions.

Additional two pins are for qvar sensing.

## Datasheet
- https://www.st.com/en/mems-and-sensors/lsm6dsv16x.html


edg/parts/Lsm6dsv16x.py
```diff
@@ -0,0 +1,82 @@
...
+        dio_model = DigitalBidir.from_supply(
+            self.gnd, self.vddio,
+            voltage_limit_abs=(-0.3*Volt, self.vddio.voltage_limits.upper()+0.3),
+            # datasheet states abs volt to be [0.3, VDD_IO+0.3], likely a typo
+            current_limits=(-4, 4)*mAmp,
+            input_threshold_factor=(0.3, 0.7)
+        )
+        self.i2c = self.Port(I2cTarget(dio_model))
+
+        dout_model = DigitalSource.low_from_supply(self.gnd)
```
ducky64, line 26: It seems the chip can do either push-pull or open-drain, though there isn't a good way to model this. Might be worth writing this in the port docs, that the port can be configurable push-pull or open-drain.

edg/parts/Lsm6dsv16x.py
```diff
@@ -0,0 +1,82 @@
...
+        self.int1 = self.Port(dout_model, optional=True)
+        self.int2 = self.Port(dout_model, optional=True)
+
+        qvar_model = DigitalSink(voltage_limits=(-0.46, 0.46)*Volt,
+                                 current_draw=(2.6, 15)*uAmp)
```
ducky64, line 31: Is qvar a digital pin? I think it's analog, maybe even touch (especially in that it should be connected to an external electrode)?

ducky64 (commented):
Mostly looks good! Just one note about current draw and a question on qvar port typing.

edg/parts/Lsm6dsv16x.py
Suke0811: So you mean we can use touch port? 

https://github.com/BerkeleyHCI/PolymorphicBlocks/blob/561da36d69b866f58d5fc6faa8a48c919b618e36/edg/electronics_model/TouchPort.py#L6-L25

edg/parts/Lsm6dsv16x.py
Suke0811: Added comments in this commit. Is this look good? 
16e55fdcb9599244685ca11b49a1e2a4c3bbac55

edg/parts/Lsm6dsv16x.py
```diff
@@ -0,0 +1,82 @@
+from ..abstract_parts import *
+from .JlcPart import JlcPart
+
+
+class Lsm6dsv16x_Device(InternalSubcircuit, FootprintBlock, JlcPart):
+    def __init__(self) -> None:
+        super().__init__()
+        self.vdd = self.Port(VoltageSink(
+            voltage_limits=(1.71, 3.6)*Volt,
+            current_draw=(4*uAmp, 0.65*mAmp)  # Values for low-power and high-performance modes
```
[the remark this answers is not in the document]
Suke0811: Updated the min value in 16e55fdcb9599244685ca11b49a1e2a4c3bbac55

edg/parts/Lsm6dsv16x.py
ducky64: I started reading more into Qvar https://www.st.com/resource/en/application_note/an5755-qvar-sensing-channel--stmicroelectronics.pdf and it seems the principle is significantly different from cap-touch that the cap-touch footprints might not be applicable. In particular, it seems Qvar needs two pins for each touch interface, while cap-touch is just one.

I think the end answer might be this should be Passive-typed for now, and in the future if we end up using Qvar a lot, it might get its own connection type.

Suke0811: image

Suke0811: d0659b546e93da5a137a06bb3fe04e7d70de230e
For now i set them to `Passive()`

ducky64 (approved):
Looks good!

One nit: consider merging the doc about push-pull / open-drain into the doc in the wrapper class. The wrapper class is going to be more visible and that doc is probably the one that will be read.

Feel free to merge whenever.

Suke0811:
Okay, I moved the doc comments to the wrapper side and merging now. Thanks!

PR #400, opened 2025-03-26
Stm32g4

Suke0811:
This will add: 
- STM32G431KB variant 
- IoControllerUsbCc in mixin
- This component will be tested on hardware very soon

The following comments from @ducky64 have been reviewed and should be applied to it 
---
- [x] PB8 is BOOT0, I think from factory it is configured to sense that pin to determine boot location. I've been handling boot pins inconsistently, but I think the argument should be for that pin to be not allocated by default, like is the case on the stm32f103 model
- [x] you have a 100uF cap for pwr_cap2, just a tad high for chip level decoupling 🤣
- [x] you might need more caps, 1x100nF per Vss/Vdd pair (3), 1x 10uF for bulk component-level decupling, 1x1uF for VddA/VssA
- [x] I2S2.ws is PF0
- [x] consider renaming dio_tt_model to dio_tta_model
- [x] analog function name for PA2 is ADC1_IN3, PA3 is ADC1_IN4 (not ADC12)
- [x] you also have the I2C3 peripheral
- [x] consider renaming the CAN peripheral 'FDCAN' to be inline with the manufacturer naming
- [x] tt IOs are rated up to Vdd+0.3 instead of an absolute 3.6 max
- [x] ftc model is rated for up to 5v absolute, the model is not shared with other IOs
- [x] DAC signal_out_bound only applies when output buffer on (probably just needs a comment)
- [x] consider adding USB peripheral mode (USP DP and USB DM)
- [x] IO pin current ratings are in table 15
- [x] DAC output impedance is spec'd between 9.6-13.8kOhm
- [x] ADC input input impedance is not the same as the external input impedance limit
- [x] for adc_model: ADCs are mostly on TT IOs, which are only tolerant up to Vdd+0.3 (aside from the FTfa and FTa ones, but I would be surprised if in ADC mode they could tolerate up to 5v)
- [x] consider adding usb power delivery support, I think the CC pins directly connect to the micro, you might need to add a IoControllerUsbCc port type mixin


edg/parts/Microcontroller_Stm32g431.py
```diff
@@ -0,0 +1,273 @@
...
+    SYSTEM_PIN_REMAP: Dict[str, Union[str, List[str]]]  # pin name in base -> pin name(s)
+    RESOURCE_PIN_REMAP: Dict[str, str]  # resource name in base -> pin name
+
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        # Power and ground
+        self.pwr = self.Port(VoltageSink(
+            voltage_limits=(1.71, 3.6) * Volt,
+            current_draw=(0.001, 150.0) * mAmp  # table 15
```
ducky64, line 25: 150mA is a max limit, not the actual draw of the device. Table 21 gives the core running current, which specifies a worst-case max of 44mA (at a toasty 125C) and low of 14nA in sleep mode (Table 32).

At some point it would probably be worth modeling these limits (total current into device pins, total IO current), but we don't currently.

edg/parts/Microcontroller_Stm32g431.py
```diff
@@ -0,0 +1,273 @@
...
+
+        adc_model = AnalogSink.from_supply(
+            self.gnd, self.pwr,
+            voltage_limit_tolerance=(-0.3, 0.3) * Volt,
+            signal_limit_tolerance=(0, 0),
+            impedance=(50, float('inf')) * kOhm  # TODO: this is maximum external input impedance, maybe restrictive
+        )
+        dac_model = AnalogSource.from_supply(
+            self.gnd, self.pwr,
+            # signal_out_bound=(0.2*Volt, -0.2*Volt), signal_out_bound only applies when output buffer on
```
ducky64, line 87: Dead code is generally discouraged style-wise. Consider just leaving the bound in there (as a safe bound) and having a comment that it only applies in a specific case.

edg/parts/Microcontroller_Stm32g431.py
```diff
@@ -0,0 +1,273 @@
...
+                'sck': ['PF1'], 'miso': ['PA10'], 'mosi': ['PA11']
+            }),
+            PeripheralFixedResource('SPI3', spi_model, {
+                'sck': ['PB3'], 'miso': ['PB4'], 'mosi': ['PB5']
+            }),
+            PeripheralFixedResource('I2S2', I2sController(DigitalBidir.empty()), {
+                'sck': ['PB13', 'PF1'], 'ws': ['PB12', 'PF0'], 'sd': ['PA11', 'PB15']
+            }),
+            PeripheralFixedResource('I2S3', I2sController(DigitalBidir.empty()), {
+                'sck': ['PB3', ], 'ws': ['PA4', 'PA15'], 'sd': ['PB5']
```
ducky64, line 145:
```suggestion
                'sck': ['PB3'], 'ws': ['PA4', 'PA15'], 'sd': ['PB5']
```
style nit

examples/EspLora/EspLora.kicad_pro
```diff
@@ -6,14 +6,14 @@
         "apply_defaults_to_fp_fields": false,
         "apply_defaults_to_fp_shapes": false,
         "apply_defaults_to_fp_text": false,
-        "board_outline_line_width": 0.0381,
+        "board_outline_line_width": 0.038099999999999995,
```
ducky64, line 9: This file shouldn't be changed since it's not relevant to this PR - can you revert it?

ducky64 (changes requested):
Overall looks good, but a few style issues and one correctness issue (related to current draw).

examples/EspLora/EspLora.kicad_pro
Suke0811: dropped this commit, hence creating a new PR

Suke0811:
Moved to #401

PR #417, opened 2025-06-30
Fix: replace the crystal to 32MHz according to the ref design

Suke0811:
https://semtech.my.salesforce.com/sfc/p/#E0000000JelG/a/RQ000008nKCH/hp2iKwMDKWl34g1D3LBf_zC7TGBRIo2ff5LMnS8r19s

This pull request includes a single change to the `contents` method in the `Rf_Sx1262.py` file. The crystal frequency was updated to align with data sheet.

* **Crystal frequency update**:
  - Changed the crystal frequency from `30 MHz` to `32 MHz` for the `Rf_Sx1262` component to align with the ref design


ducky64 (approved):
Good catch, lgtm!

